PR #68, opened 2025-07-23
Set config directory early to avoid bugs handling the empty string (those should be fixed with this change as well)

AndiDog:
### What does this PR do?

Some `if` branches in the code treated an empty config path differently which caused hard-to-trace bugs like not loading any services. This sets the config path early to avoid even having an empty value.


lyind (approved):
Did we ever have an empty value? I'm assuming some of these code location are run in go-routines or separate processes?

Would be nice to attach some log files (e.g. captured during debugging), but otherwise, I'll trust you diagnosed the root cause correctly.

JosephSalisbury (approved):

PR #65, opened 2025-07-21
Feat/systemd socket activation

lyind:
### Summary

This PR implements starting the systemd service via socket activation (aggregator socket, port 8090), which saves resources and speeds up system startup.

### Should this change be mentioned in the release notes?

- [x] CHANGELOG.md has been updated (if it exists)

PR #72, opened 2025-07-23
Feat/systemd socket activation

lyind:
Follows up https://github.com/giantswarm/muster/pull/65

### What does this PR do?

Fix multiple listeners use-case required for out-of-the-box platform compatibility.

Muster works on all platforms after running setup scripts, without systemd socket listener configuration required.

- no, it's just a fix for the linked PR

### DIFF
--- a/internal/aggregator/server.go
+++ b/internal/aggregator/server.go
@@ -3,13 +3,15 @@ package aggregator
 import (
 	"context"
 	"fmt"
-	"muster/internal/config"
-	"muster/pkg/logging"
+	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


teemow:
Sorry. But I don't think this a good idea. This would mean that by default Muster will start with two separate http servers within the same process. This is not only a waste of memory but its also totally unclear to me how state will be handled and can lead to all kind of problems.

I already suggested how to fix this in the other PR. Depending on their system setup people need to configure the socket correctly. And by default ipv4 should be fine imo.

PR #87, opened 2025-07-24
Increases parallel execution and decreases times generally

JosephSalisbury:
- Increases number of workers for tests running in parallel
- Generally removes / decreases waits and timeouts

Changes test execution from ~1m30 to ~20s


teemow (approved):
❤️

PR #92, opened 2025-07-24
Removes SBOM and sign options from goreleaser dry-run and add cache for go build

JosephSalisbury:
- Removes unnecessary goreleaser steps
- Adds cache for golang build cache (safe between builds)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -23,6 +23,16 @@ jobs:
         with:
           go-version: '1.24'  # Match the version used in auto-release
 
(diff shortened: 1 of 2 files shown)


ubergesundheit:
if this repo gets aligned to how other gs repos are built, goreleaser will go away

lyind:
> if this repo gets aligned to how other gs repos are built, goreleaser will go away

I'm of the opinion we should do this within the scope of this hackathon. Discussing this with Joe.

mproffitt (commented):
LGTM

PR #95, opened 2025-07-24
add Dockerfile and edit circleci config

QuantumEnigmaa:
### What does this PR do?

This PR adds a Dockerfile for allowing muster to be deployed as an app in a cluster thanks to a follow-up PR that will add a Helm chart. 

It also updates the circleci config so that muster is pushed to our image registries.

### Testing

- [ ] For UI changes involving golden files (e.g., in `internal/tui/view/testdata/`), new files have been generated/updated (e.g., by running `go test ./internal/tui/view/... -update`) and meticulously verified.

- [ ] CHANGELOG.md has been updated (if it exists)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,22 +8,22 @@ workflows:
     jobs:
       - architect/go-build:
           context: architect
-          name: go-build
-          binary: template
+          name: go-build-muster
+          binary: muster
           filters:
             # Trigger job also on git tag.
             tags:
               only: /^v.*/
-      # - architect/push-to-registries:
-      #     context: architect
-      #     name: push-to-registries
-      #     requires:
-      #       - go-build
-      #     filters:
-      #       # Trigger job also on git tag.
-      #       tags:
-      #         only: /^v.*/
-      #       branches:
-      #         ignore:
-      #           - main
-      #           - master
+
+      - architect/push-to-registries:
+          context: architect
+          name: push-to-registries
+          requires:
+          - go-build-muster
+          filters:
+            tags:
+              only: /^v.*/
+            branches:
+              ignore:
+                - main
+                - master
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -8,22 +8,22 @@ workflows:
     jobs:
       - architect/go-build:
           context: architect
-          name: go-build
-          binary: template
+          name: go-build-muster
```
QuentinBisson, line 11:
```suggestion
          name: go-build
```

.circleci/config.yml
```diff
@@ -8,22 +8,22 @@ workflows:
...
-      #       branches:
-      #         ignore:
-      #           - main
-      #           - master
+
+      - architect/push-to-registries:
+          context: architect
+          name: push-to-registries
+          requires:
+          - go-build-muster
```
QuentinBisson, line 22:
```suggestion
          - go-build
```

.circleci/config.yml
QuantumEnigmaa: Don't you think the name makes more sense as it is currently ?

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM gsoci.azurecr.io/giantswarm/alpine:3.20.3-giantswarm
```
QuentinBisson, line 1: What's the use of those 2 from btw?

QuantumEnigmaa:
The last commit is an attempt to fix the current issue encountered when trying to run a container from the image generated by the previous version of the Dockerfile:
```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during contain
er init: exec: "sh": executable file not found in $PATH: unknown
```
Building from `scratch` is not enough for the muster binary.

But we'll have to update the base image that is currently `gcr.io/distroless/base-debian12:nonroot`

PR #99, opened 2025-07-24
fix default config: AggregatorConfig.Enabled=true

lyind:
Resolves https://github.com/giantswarm/giantswarm/issues/33976

### What does this PR do?

- Set `AggregatorConfig.Enabled: true` in default constructed configs, to better support running with incomplete/empty/default configuration. 
- ~~(Add missing config field overrides (probably forgotten when these fields were added) )~~
  - Removed legacy code function `mergeConfig()`
  - Replaced redundant and racy `os.Stat` calls (to detect non-existent config) completely
- Remove `AggregatorConfig.Enabled` field (always enabled where it's required)
- Refactor and unify: `DetectAggregatorEndpoint...` -> `GetAggregatorEndpoint()`
- Fix previously unnoticed coding error where `agentConfigPath` wasn't used
- Refactor `LoadConfig()` to always expect a path

- `AggregatorConfig.Enabled` field dropped from `config.yaml`
- No problems running `giantswarm/debug` or `muster` out-of-the-box.
- `config-path` is honored for endpoint selection in `muster agent` modes

Pure bugfix and on-boarding enhancement, no changelog entry needed.

### DIFF
--- a/.muster/config.yaml.example
+++ b/.muster/config.yaml.example
@@ -6,5 +6,4 @@
 aggregator:
   port: 8090
   host: localhost
-  # enabled: true
   # musterPrefix: "x"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,8 @@ All notable changes to this project will be documented in this file.
   - Documented dependency management, state management, and message flow in detail
 
 ### Changed
+- **Aggregator Config**
+  - Drop the "Enabled" field (always enabled in modes where it's used)
 - **Service Manager Refactoring**
   - ServiceManager now accepts an optional KubeManager parameter for K8s connection services
   - Added support for K8s connection services in the service lifecycle management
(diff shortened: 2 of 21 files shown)


teemow:
FYI @lyind the merge config comes from envctl where I supported both a project level and a user level config. This has been removed in muster. Could be that this function is a leftover that can be refactored/removed too to make the config loading simpler.

lyind:
@teemow Is the following assumption correct?

> `muster serve` must always run the mcp-aggregator service.

teemow:
Thinking about it. How about we remove this config flag completly? I don't see any use case anymore where you would start `muster serve` without the aggregator. This is legacy code configuration as well.

lyind:
I think the idea and management of "services" is generic. All services have an `Enabled` flag. It doesn't seem bad to be able to switch stuff off without removing the service config.

lyind:
I have rewritten and tested (with config file and without) the `LoadConfigFromPath()` code. Should be "good enough". ;)

lyind:
I had the impression `AggregatorConfig` was reused for something else. However, that doesn't appear to be the case.

Removed all mention of `AggregatorConfig.Enabled` (always true behavior).

JosephSalisbury (approved):

PR #103, opened 2025-07-25
make muster CRs discovery namespace configurable

QuantumEnigmaa:
### What does this PR do?

This PR makes the discovery namespace for CRs configurable. This is needed to make muster work in a k8s environment.

### DIFF
--- a/internal/app/services.go
+++ b/internal/app/services.go
@@ -109,7 +109,7 @@ func InitializeServices(cfg *Config) (*Services, error) {
 
 	// Step 1: Create unified muster client once
 	// This avoids redundant Kubernetes connection attempts and CRD validation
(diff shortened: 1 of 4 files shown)


internal/app/services.go
```diff
@@ -109,7 +109,7 @@ func InitializeServices(cfg *Config) (*Services, error) {
 
 	// Step 1: Create unified muster client once
 	// This avoids redundant Kubernetes connection attempts and CRD validation
-	musterClient, err := createMusterClient(cfg.ConfigPath, cfg.Debug)
+	musterClient, err := createMusterClientWithConfig(cfg.ConfigPath, cfg.Debug, *cfg.MusterConfig)
```
teemow, line 112: This is a typical problem with the agent. Please make sure there is only one way to create the client and not multiple.

teemow (changes requested):

PR #104, opened 2025-07-25
client: fail if endpoints do not match

lyind:
### What does this PR do?

Fail client executions if the endpoint configured in systemd user socket doesn't match `config.yaml`.

User will know why his `muster` client can't connect to the muster launched via systemd: `config.yaml` misconfigured.

### How does it look like?

The error message will look like this (positive/negative tests performed):
```
Error: systemd muster.socket doesn't match configured endpoint http://localhost:8099/mcp. Please check your configuration in /home/lyind/.config/muster

Process finished with the exit code 1
```

No CHANGELOG.md entry needed, because pure error avoidance/usability improvement.

### DIFF
--- a/internal/cli/common.go
+++ b/internal/cli/common.go
@@ -3,6 +3,9 @@ package cli
 import (
 	"fmt"
 	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	"muster/internal/config"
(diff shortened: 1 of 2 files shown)


lyind:
Not sure if we should also check for this on the `muster serve` side (it technically may not matter there)?

lyind:
@teemow because we discussed this recently:

This PR is intended to reduce friction for users who run **both** client and server directly on their local system (not in a container) and opt to launch the `muster` server on demand, via systemd socket activation (more common on some Linux distributions or Windows WSL2).

Which might affect many developers and people who try things out.

Other use cases should not be affected.

teemow:
My plan is to remove the local support except running it in kind - making kubernetes the only deployment target to maintain.

PR #438, opened 2026-03-12
refactor: replace X-Muster-Session-ID with mcp-go native Mcp-Session-Id

gusevda:
### What does this PR do?

Migrates session management from a custom `X-Muster-Session-ID` header to mcp-go's built-in `Mcp-Session-Id` mechanism, as defined in the [MCP spec — Session Management](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management). This replaces the hand-rolled `clientSessionIDMiddleware` with mcp-go's `SessionIdManager` interface (`MusterSessionIdManager`), uses `WithHTTPContextFunc` for identity binding, and `WithHooks` (OnAfterInitialize) for proactive SSO.

Key changes:
- **New `MusterSessionIdManager`** — implements `Generate`, `Validate`, `Terminate` backed by the existing `SessionRegistry`
- **Removed `clientSessionIDMiddleware`** — session lifecycle is now handled by mcp-go internally
- **Added `httpContextFunc`** — binds the authenticated user's subject to the mcp-go session (replaces identity binding from the old middleware)
- **Added `AfterInitialize` hook** — triggers proactive SSO when a client completes the MCP handshake
- **Header rename** — `X-Muster-Session-ID` → `Mcp-Session-Id` throughout (client, server, tests)
- **Removed `GetClientSessionIDFromContext` / `WithClientSessionID`** — no longer needed since mcp-go manages session context natively

CLI clients will now use the standard `Mcp-Session-Id` header instead of the custom `X-Muster-Session-ID` header. Session persistence across CLI invocations continues to work — the client intercepts the `Mcp-Session-Id` response header and replays it on subsequent requests.

The custom `X-Muster-Session-ID` header was introduced before mcp-go had native session management support. Now that mcp-go provides `SessionIdManager` and `WithHTTPContextFunc`, we can delegate session lifecycle to the library and remove ~50 lines of custom middleware.

- Verify the `MusterSessionIdManager` correctly bridges the `SessionRegistry` into mcp-go's lifecycle
- Confirm identity binding via `httpContextFunc` is equivalent to the old middleware behavior
- Check that proactive SSO via `AfterInitialize` hook fires at the right time

### Testing

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- Replaced custom `X-Muster-Session-ID` header with mcp-go's native `Mcp-Session-Id` header for session tracking. The `MusterSessionIdManager` middleware now handles session ID assignment using the standard MCP header.
+
 ## [0.1.0] - 2026-02-23
 
 ### Changed
--- a/internal/agent/client.go
+++ b/internal/agent/client.go
@@ -69,7 +69,7 @@ type Client struct {
 	cacheEnabled      bool
 	formatters        *Formatters
 	NotificationChan  chan mcp.JSONRPCNotification
-	headers           map[string]string           // Custom HTTP headers (e.g., X-Muster-Session-ID)
+	headers           map[string]string           // Custom HTTP headers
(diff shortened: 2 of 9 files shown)


teemow:
@gusevda these concepts are fundamentally different and imo mcp-gos session manager is not a replacement for the session management we need inside muster.

PR #543, opened 2026-03-25
Deduplicate aggregated MCP tools from servers exposing the same tools

TheoBrigitte:
## Feature

When multiple backend MCP servers expose tools with the same original name, the aggregator now automatically deduplicates them into a single exposed tool with a required `server` parameter. The deduplicated tool uses only the muster prefix and the original tool name — no server name in the tool name.

**Before** — two kubernetes MCP servers (`kubernetes-graveler`, `kubernetes-gazelle`) both exposing `kubernetes_api_resources` produced duplicate tools:

```
x_kubernetes-graveler_kubernetes_api_resources
x_kubernetes-graveler_kubernetes_apply
x_kubernetes-graveler_kubernetes_get
...
x_kubernetes-gazelle_kubernetes_api_resources
x_kubernetes-gazelle_kubernetes_apply
x_kubernetes-gazelle_kubernetes_get
...
```

(38 tools — every tool duplicated)

**After** — the aggregator detects that `kubernetes_api_resources` exists on multiple servers and exposes a single deduplicated tool:

```
x_kubernetes_api_resources
x_kubernetes_apply
x_kubernetes_get
x_list_port_forward_sessions
...
```

(19 tools — each with a required `server` parameter)

Each deduplicated tool has a required `server` enum parameter listing the available backends:

```json
{
  "name": "x_kubernetes_api_resources",
  "description": "... (available on servers: kubernetes-gazelle, kubernetes-graveler)",
  "inputSchema": {
    "properties": {
      "server": {
        "type": "string",
        "description": "Target server to execute this tool on. Available: kubernetes-gazelle, kubernetes-graveler",
        "enum": ["kubernetes-gazelle", "kubernetes-graveler"]
      }
    },
    "required": ["server"]
  }
}
```

Tools unique to a single server continue to use the normal prefixed naming: `x_{serverName}_{toolName}`.

### Unit Tests

```
$ go test ./internal/aggregator/ -run TestAggregatorServer_HandlerTracking -v
=== RUN   TestAggregatorServer_HandlerTracking
--- PASS: TestAggregatorServer_HandlerTracking (0.20s)
PASS

$ go test ./internal/aggregator/ -run TestServerRegistry -v
=== RUN   TestServerRegistry_AlwaysPrefixing
=== RUN   TestServerRegistry_AlwaysPrefixing/All_tools_get_prefixed
=== RUN   TestServerRegistry_AlwaysPrefixing/Tools_with_same_names_get_prefixed
=== RUN   TestServerRegistry_AlwaysPrefixing/Multiple_servers_with_same_tool
--- PASS: TestServerRegistry_AlwaysPrefixing (0.00s)
=== RUN   TestServerRegistry_ResolveName
=== RUN   TestServerRegistry_ResolveName/x_serverA_unique_tool
=== RUN   TestServerRegistry_ResolveName/x_serverA_shared_tool
=== RUN   TestServerRegistry_ResolveName/x_serverB_shared_tool
=== RUN   TestServerRegistry_ResolveName/x_serverA_unique_prompt
=== RUN   TestServerRegistry_ResolveName/x_serverB_shared_prompt
=== RUN   TestServerRegistry_ResolveName/x_serverC_shared_prompt
=== RUN   TestServerRegistry_ResolveName/non_existent
--- PASS: TestServerRegistry_ResolveName (0.00s)
PASS
```

### Integration Test

With two MCP servers both exposing the same tools:

1. **Tool listing** — `list_tools` returns deduplicated tools instead of duplicates
2. **Calling without `server`** — Returns an error: `tool x_kubernetes_api_resources is available on multiple servers (kubernetes-gazelle, kubernetes-graveler); the 'server' parameter is required`
3. **Calling with `server`** — Routes to the correct backend, stripping the `server` param before forwarding
4. **Server deregistration** — When a server is removed, its tools revert to normal prefixed naming on the remaining server

## Changes

- **`internal/aggregator/registry.go`** — Rewrote `GetAllTools` and `GetAllToolsForSession` with a two-pass algorithm: Pass 1 groups tools by their original backend name across all connected servers; Pass 2 builds the exposed list where single-server tools get prefixed normally (`x_{serverName}_{toolName}`) and multi-server tools are deduplicated as `x_{originalToolName}` with a required `server` enum parameter. Added `registerDeduplicatedTool` for reverse-lookup mappings. Made `server` a required field in `addServerParam`.
- **`internal/aggregator/server.go`** — Updated `CallToolInternal` to enforce the required `server` parameter for multi-server tools (returns an error if omitted instead of routing to a default). Added explicit-server resolution path and auth short-circuit for deduplicated tools targeting OAuth servers.
- **`internal/aggregator/server_test.go`** — Updated `TestAggregatorServer_HandlerTracking` to expect 3 tools (deduplicated `x_shared-tool`) instead of 4 (duplicate `x_server1_shared-tool` + `x_server2_shared-tool`).
- **`CHANGELOG.md`** — Added changelog entry.

### DIFF
--- a/internal/aggregator/registry.go
+++ b/internal/aggregator/registry.go
@@ -3,6 +3,7 @@ package aggregator
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -50,6 +51,12 @@ type ServerRegistry struct {
 	serverPrefixes map[string]string // server name -> configured prefix
 	nameMu         sync.RWMutex
 	musterPrefix   string
+
+	// allMappings tracks all servers that provide each exposed name.
+	// When multiple servers share the same toolPrefix and expose the same tool,
+	// their exposed names collide. This map records every server for deduplication
+	// and multi-server routing (protected by nameMu).
+	allMappings map[string][]resolvedName
 }
 
 // NewServerRegistry creates a new server registry with the specified global prefix.
@@ -71,6 +78,7 @@ func NewServerRegistry(musterPrefix string) *ServerRegistry {
 		nameMapping:    make(map[string]resolvedName),
 		serverPrefixes: make(map[string]string),
 		musterPrefix:   musterPrefix,
+		allMappings:    make(map[string][]resolvedName),
 	}
 }
 
@@ -82,7 +90,20 @@ func (r *ServerRegistry) ExposedToolName(serverName, toolName string) string {
 	r.nameMu.Lock()
 	defer r.nameMu.Unlock()
 	exposed := r.buildExposedNameLocked(serverName, toolName)
-	r.nameMapping[exposed] = resolvedName{serverName: serverName, originalName: toolName, itemType: "tool"}
+	rn := resolvedName{serverName: serverName, originalName: toolName, itemType: "tool"}
+	r.nameMapping[exposed] = rn
+	// Track all servers providing this exposed name for deduplication.
+	existing := r.allMappings[exposed]
+	alreadyTracked := false
+	for _, m := range existing {
+		if m.serverName == serverName {
+			alreadyTracked = true
+			break
+		}
+	}
+	if !alreadyTracked {
+		r.allMappings[exposed] = append(existing, rn)
+	}
 	return exposed
 }
 
@@ -289,6 +310,10 @@ func (r *ServerRegistry) GetAllTools() []mcp.Tool {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
 
+	// seen tracks exposed name → index in allTools for deduplication.
+	seen := make(map[string]int)
+	// serversByTool tracks which servers provide each exposed tool name.
+	serversByTool := make(map[string][]string)
 	var allTools []mcp.Tool
 	connectedCount := 0
 	authRequiredCount := 0
@@ -317,11 +342,27 @@ func (r *ServerRegistry) GetAllTools() []mcp.Tool {
 			// Apply smart prefixing to avoid name conflicts
 			exposedTool := tool
 			exposedTool.Name = r.ExposedToolName(serverName, tool.Name)
-			allTools = append(allTools, exposedTool)
+
+			if _, exists := seen[exposedTool.Name]; !exists {
+				seen[exposedTool.Name] = len(allTools)
+				serversByTool[exposedTool.Name] = []string{serverName}
+				allTools = append(allTools, exposedTool)
+			} else {
+				serversByTool[exposedTool.Name] = append(serversByTool[exposedTool.Name], serverName)
+			}
 		}
 		info.mu.RUnlock()
 
 		logging.Debug("Aggregator", "Server %s has %d tools", serverName, serverToolCount)
+	}
+
+	// For deduplicated tools backed by multiple servers, inject a "server" parameter.
+	for exposedName, servers := range serversByTool {
+		if len(servers) <= 1 {
+			continue
+		}
+		idx := seen[exposedName]
+		allTools[idx] = addServerParam(allTools[idx], servers)
 	}
 
 	logging.Debug("Aggregator", "GetAllTools: returning %d tools from %d connected servers (%d servers require auth, use core_auth_login)",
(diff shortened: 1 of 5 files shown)


teemow (changes requested):
Thanks @TheoBrigitte for pushing this forward. But I think it would be better to configure this explicitly in the MCP server CRs. The magic might be surprising for users.

Scenario:
* Servers A and B both expose foo - deduped as x_foo
* Server A is deregistered
* GetAllTools now exposes x_serverB_foo (single-server prefix)
* CallToolInternal("x_foo", ...) still resolves via nameMapping to server A (which no longer exists), producing a "server not found" error
* GetToolServerNames("x_foo") still returns both [serverA, serverB]

Let's make sure we configure a stable interface that is not depending on how many servers there are. Also we have customers that only have one server and we should build the skills around the same tool names.

marians:
I totally agree that this needs improvement and have a less impactful PR here: https://github.com/giantswarm/muster/pull/570

QuentinBisson:
@TheoBrigitte and @teemow what do you think about adding `spec.family` to declare which family each backend is an instance of? I think it works better than group?

QuentinBisson:
@TheoBrigitte do you want me to take this PR over?

internal/aggregator/registry.go
```diff
@@ -312,16 +323,36 @@ func (r *ServerRegistry) GetAllTools() []mcp.Tool {
...
 		for _, tool := range info.Tools {
-			// Apply smart prefixing to avoid name conflicts
-			exposedTool := tool
-			exposedTool.Name = r.ExposedToolName(serverName, tool.Name)
-			allTools = append(allTools, exposedTool)
+			toolsByOrigName[tool.Name] = append(toolsByOrigName[tool.Name], toolEntry{
+				serverName: serverName,
+				tool:       tool,
+			})
 		}
```
QuentinBisson, line 331: Dedup keys on `tool.Name` only. Two backends advertising the same tool name with different descriptions (or schemas) get silently merged — one wins arbitrarily. Worth keying on `(name, description)` at minimum so a backend running an older/different version doesn't get its semantics erased. Same shape duplicated at L719.

internal/aggregator/registry.go
```diff
@@ -789,3 +877,53 @@ func (r *ServerRegistry) IsOAuthServer(serverName string) bool {
...
+	}
+
+	// Make "server" a required parameter.
+	tool.InputSchema.Required = append(tool.InputSchema.Required, "server")
+
+	// Update description to note multi-server availability.
+	tool.Description = tool.Description + " (available on servers: " + strings.Join(sorted, ", ") + ")"
+
+	return tool
+}
```
QuentinBisson, line 929: `tool` is passed by value but `InputSchema.Properties` is a map and `InputSchema.Required` is a slice — both reference types that share storage with the originating `info.Tools[i]`. The map write on L916 and the `append` on L923 mutate the backend's stored copy. Subsequent `GetAllTools` calls will see the injected `server` field (and possibly `"server"` in `Required`) on every server's stored tool. Need to deep-copy `Properties` and start a fresh `Required` slice before mutation.

internal/aggregator/registry.go
```diff
@@ -789,3 +877,53 @@ func (r *ServerRegistry) IsOAuthServer(serverName string) bool {
...
+	mappings := make([]resolvedName, len(serverNames))
+	for i, sn := range serverNames {
+		mappings[i] = resolvedName{serverName: sn, originalName: originalName, itemType: "tool"}
+	}
+	r.allMappings[exposedName] = mappings
+	// Set nameMapping to first entry so ResolveToolName can still find it.
+	if len(mappings) > 0 {
+		r.nameMapping[exposedName] = mappings[0]
+	}
+}
```
QuentinBisson, line 897: No corresponding cleanup path. `Deregister()` at L235 only removes from `r.servers`, leaving `r.allMappings` and `r.nameMapping` populated. After `{A, B} expose foo → A is deregistered`, `r.allMappings["x_foo"]` still contains `[A, B]` and `r.nameMapping["x_foo"]` still points at A. `CallToolInternal("x_foo", server: "serverA")` then routes to a deregistered server and surfaces as "server not found" rather than "no such tool." Cleanup needs to walk both maps and drop entries owned by the deregistering server.

internal/aggregator/server.go
```diff
@@ -1699,7 +1700,45 @@ func (a *AggregatorServer) CallToolInternal(ctx context.Context, toolName string
...
+		// Resolve against the explicitly requested server.
+		originalName, err = a.registry.ResolveToolNameForServer(toolName, targetServer)
+		if err == nil {
+			serverName = targetServer
+		}
+	}
+	if serverName == "" {
+		// Fall back to default resolution (single-server tools).
+		serverName, originalName, err = a.registry.ResolveToolName(toolName)
+	}
```
QuentinBisson, line 1740: When `targetServer` is set but `ResolveToolNameForServer` errors (e.g. caller passed `server: "typo"`), `serverName` stays empty and the code falls through to `ResolveToolName(toolName)`, routing to whichever backend the legacy `nameMapping` happens to point at. An explicit-but-invalid `server` arg should return the resolution error directly, not silently re-route.

QuentinBisson (commented):
Bugs flagged inline. Design feedback in a separate thread.

QuentinBisson:
Closing in favor of #670, which reworks this from automatic deduplication into an explicit `MCPServer.spec.family` opt-in. This addresses @teemow's CHANGES_REQUESTED review (stable interface that doesn't depend on instance count, single-server skills work the same as multi-server) and fixes the four inline bugs flagged from this review (dedup key, deep-copy of input schema, Deregister cleanup, explicit-server error path).

Thanks @TheoBrigitte for pushing the original direction — the dedup machinery and `server` enum approach carry over.

Follow-up: #668 tracks renaming `toolPrefix` once `family` is the canonical naming source.

PR #566, opened 2026-04-17
feat: accept ID tokens issued directly by the upstream OIDC provider

paurosello:
## Summary

- Adds `trustedAudiences` to `OAuthServerConfig`, wired through to `mcp-oauth`'s `server.Config` (and exposed through the Helm chart), so operators can opt in to JWKS-validated cross-client token acceptance.
- When `ValidateToken` succeeds but muster's token store has no entry for the bearer (direct-IdP case), treat the bearer JWT as the ID token: synthesize a deterministic `ext-<sha256(token)[:16]>` session ID, inject subject / sessionID / idToken into the request context, and mirror the ID token into the OAuth proxy store keyed by the provider's issuer URL (Dex issuer for Dex provider, `baseURL` otherwise). The existing `onAuthenticated` hook then runs `initSSOForSession` and populates `authStore` / `capabilityStore` on first use.

## Motivation

Currently, muster only accepts bearer tokens it issued itself. Presenting a JWT obtained directly from Dex (e.g. via `dex-k8s-authenticator` or another trusted Dex client) passes mcp-oauth's `TrustedAudiences` validation but leaves the request with no session ID — the first tool call then fails with `user not authenticated to server X` because `authStore` has no entry for the synthetic session.

## Example config

```yaml
aggregator:
  oauth:
    server:
      enabled: true
      provider: dex
      dex:
        issuerUrl: https://dex.example.com
        clientId: muster
        clientSecret: ...
      trustedAudiences:
        - dex-k8s-authenticator
        - some-other-dex-client
```

## Unit test plan

- [x] `go test ./internal/server/...` — `TestInjectExternalIDToken` covers: non-JWT rejection, JWT without derivable subject, happy-path injection, `userInfo.ID` vs JWT `sub` precedence, session-ID determinism, `onAuthenticated` invocation, issuer-key resolution for Dex vs non-Dex providers, handler-disabled, missing issuer.
- [x] `go test ./internal/config/...`
- [x] `go build ./...`
- [x] Helm chart renders `oauth.server.trustedAudiences` into `config.yaml`.

## End-to-end on graveler

Validated on graveler via a throwaway branch of `giantswarm-management-clusters` pinning the OCIRepository to this PR's chart build. Dex ID token (client `gscliauth`) accepted, session `ext-...` synthesized, 16 downstream MCP kubernetes tools registered, `tools/list` returned successfully.

## Notes / open question

- **PRIORITY 2 fallback**: mcp-oauth also accepts external tokens via the provider's userinfo endpoint when the JWKS+audience path doesn't match. This means tokens whose `aud` is *not* in `trustedAudiences` still get in as long as the provider validates them. We may want to tighten `injectExternalIDToken` to only fire when `userInfo.TokenSource == TokenSourceSSO` (i.e. the JWKS path), which would restrict external-token acceptance strictly to the configured audience allow-list. Out of scope for this PR.
- `ext-` prefixed session IDs distinguish synthetic sessions from muster-issued token family IDs in logs / dashboards.
- The session ID is derived deterministically from the bearer token, so requests with the same token share cache entries; different tokens — including rotations from the same user — get independent sessions (same as the behavior today for muster-issued tokens keyed by family ID).


paurosello:
### How `trustedAudiences` fits with mcp-oauth's two validation paths

For background, mcp-oauth's `ValidateToken` tries two paths, in order:

**PRIORITY 1 — `TrustedAudiences` + JWKS** ([flows.go](https://github.com/giantswarm/mcp-oauth/blob/v0.2.101/server/flows.go#L265))

Fires only when `TrustedAudiences` is non-empty and the bearer is a JWT. mcp-oauth parses the JWT's `aud` claim; if it matches one of `TrustedAudiences`, it fetches the provider's JWKS and cryptographically verifies the signature, issuer, and audience locally. Returns `userInfo` with `TokenSource == TokenSourceSSO`.

- **Local** — no network call to the provider.
- **Strong guarantee** — signature + issuer + aud are cryptographically checked.
- **Explicit allow-list** — only clients you named get in.

**PRIORITY 2 — token store + provider userinfo** ([flows.go](https://github.com/giantswarm/mcp-oauth/blob/v0.2.101/server/flows.go#L285))

The fallback. Two sub-cases:

1. *muster-issued token*: `validateStoredToken` finds it in muster's OAuth token store and looks up the stored upstream provider token. `provider.ValidateToken` is then called with that upstream token against the provider's userinfo endpoint.
2. *external token not in the store*: `validateStoredToken` returns `(nil, nil)`. `provider.ValidateToken` is called with the raw bearer against the provider's userinfo endpoint. If the provider accepts it (graveler's Dex userinfo happens to accept ID tokens), mcp-oauth returns `userInfo` with `TokenSource == TokenSourceOAuth`.

### Why `trustedAudiences` exists even though PRIORITY 2 "already works"

In the graveler e2e test the Dex-direct path succeeded even with a token whose `aud` wasn't in `trustedAudiences` — mcp-oauth fell through to PRIORITY 2, Dex's userinfo accepted the ID token, and my middleware took over from there. So why bother configuring `trustedAudiences`?

1. **Performance.** PRIORITY 1 is fully local. PRIORITY 2 does a userinfo round-trip to the IdP on every single MCP request. At scale that's one extra RTT per tool call per user.
2. **Portability.** Dex's userinfo accepting ID tokens is implementation-specific behavior (the OIDC spec actually requires an access token at userinfo). Some providers reject ID tokens there outright — direct-IdP acceptance would stop working the moment you point muster at one.
3. **Stronger validation.** PRIORITY 1 cryptographically verifies the signature, expected issuer, and audience against the provider's JWKS. PRIORITY 2 trusts whatever the provider's userinfo returns.
4. **Explicit trust.** `aud` is the JWT claim for "this token is intended for X". Without `trustedAudiences`, muster accepts *any* token the provider's userinfo endpoint validates — effectively "anyone who can log in to your Dex gets muster access, regardless of which client issued their token." With `trustedAudiences`, muster declares "I only accept tokens minted for these specific clients."
5. **Operational clarity.** Log line at startup says exactly which audiences are trusted; there's no ambient behavior to audit.

### Open question: should we tighten PRIORITY 2 for external tokens?

Today `injectExternalIDToken` fires whenever `getProviderToken` returns nil — which covers *both* TrustedAudiences-validated tokens *and* userinfo-fallback-validated external tokens. We could restrict it to the former by gating on `userInfo.TokenSource == TokenSourceSSO`. That would make `trustedAudiences` the single source of truth for which external tokens are accepted, and turn the PRIORITY 2 userinfo path back into "muster-issued tokens only." Happy to do that in a follow-up if we want the stricter posture.

paurosello:
## Testing locally from the terminal

This reproduces end-to-end against any reachable Dex. We used graveler's Dex; substitute your own.

**1. Run muster locally** pointed at a real Dex. OAuth server enabled, `provider: dex`, `trustedAudiences` populated with the Dex `client_id`s you want to accept tokens from.

```bash
go install && muster serve --config-path ./local-config
```

**2. Obtain a Dex ID token via OAuth 2.0 device code flow** for one of the clients in `trustedAudiences`. Device flow avoids any browser redirect setup.

```bash
DEX=https://dex.example.com
CLIENT_ID=my-trusted-client

# Start device authorization. Optionally pin a connector with connector_id=...
curl -s -X POST $DEX/device/code \
  -d "client_id=$CLIENT_ID" \
  -d "scope=openid profile email groups"
# -> { "device_code": "...", "user_code": "ABCD-EFGH",
#      "verification_uri_complete": "https://dex.example.com/device?user_code=ABCD-EFGH", ... }
```

Visit the `verification_uri_complete` and approve, then poll the token endpoint with the returned `device_code`:

```bash
curl -s -X POST $DEX/token \
  -d "grant_type=urn:ietf:params:oauth:grant-type:device_code" \
  -d "device_code=<device_code>" \
  -d "client_id=$CLIENT_ID"
# -> { "access_token": "...", "id_token": "eyJ...", ... }
```

**3. Inspect the `aud` claim** to confirm it matches one of the entries in `trustedAudiences`:

```bash
ID_TOKEN=eyJ...   # from step 2
python3 -c "import json,base64,sys; p=sys.argv[1].split('.')[1]; print(json.loads(base64.urlsafe_b64decode(p+'='*(-len(p)%4))))" "$ID_TOKEN"
```

**4. Initialize an MCP session with the ID token as bearer**:

```bash
curl -sD /tmp/h.txt -X POST http://localhost:8090/mcp \
  -H "Authorization: Bearer $ID_TOKEN" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"test","version":"1"}}}'
SESSION=$(grep -i mcp-session-id /tmp/h.txt | awk '{print $2}' | tr -d '\r')
```

Expected: `HTTP/2 200` with a result object. Muster's logs should show:
```
SSO: onAuthenticated callback sessionID=ext-... hasIDToken=true
SSO: initSSOForSession called (...)
User ... connected to <server> via SSO token forwarding with N tools
```

**5. Call a tool** to confirm SSO token forwarding works end-to-end:

```bash
curl -s -X POST http://localhost:8090/mcp \
  -H "Authorization: Bearer $ID_TOKEN" \
  -H "Content-Type: application/json" \
  -H "Accept: application/json, text/event-stream" \
  -H "mcp-session-id: $SESSION" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'
```

Expected: the tool list includes any SSO-forwarded downstream servers. No `no ID token in OAuth store` warnings in the logs.

paurosello:
Another way to test this:

- login into happa of the MC
- use browser inspector the get berer token
- start mcp inspector `npx @modelcontextprotocol/inspector`
- setup auth with `Bearer <happa token>`
- use muster endpoint `https://muster.graveler.gaws2.gigantic.io/mcp`

Execute List Tools, or something else like x_graveler-mcp-kubernetes_capi_list_clusters

QuentinBisson (approved):

paurosello:
## Guidance: intermediary / token-broker apps (e.g. bedrock-gateway style)

Follow-up question came up: for customers running an intermediary app that "logs in" to multiple providers and injects bearer tokens on outbound calls — should that app authenticate against **muster** or against **Dex directly**?

**Short answer: Dex directly.** PR #566 is the feature that makes this clean; use it.

### Why Dex-direct fits the intermediary pattern

The intermediary already speaks OIDC to multiple providers and holds refresh logic. Putting muster in front of Dex means configuring muster as Yet Another Provider, storing another refresh token, and handling another `/token` endpoint — all to end up back at Dex on the backend. Pure overhead, no policy benefit.

### Concrete differences

| Concern | Login via muster | Login via Dex directly |
|---|---|---|
| Token the client holds | muster access + muster refresh | Dex `id_token` + Dex refresh (`offline_access`) |
| Refresh endpoint | muster `/token` | Dex `/token` |
| Session state in muster | Persistent, OAuth-store-backed | Synthetic (`ext-<sha256(token)[:16]>`), stateless |
| Token validity | Muster config | Dex config (`id_token` TTL + refresh rotation) |
| Portability to other backends | No — muster-only | Yes — any service that trusts Dex |
| Per-request cost on muster | Token store lookup | PRIORITY 1: local JWKS verify when audience is in `trustedAudiences` |

### Refresh & expiration

**Dex-direct flow:**
1. Intermediary requests scopes `openid profile email groups offline_access` from Dex.
2. Intermediary persists the Dex refresh token and rotates it proactively (refresh at ~80% of TTL, not reactively on 401 — otherwise in-flight requests race the expiry).
3. On every outbound call to muster, inject `Authorization: Bearer <dex id_token>`.
4. Each refresh produces a fresh JWT → the synthetic session ID in muster changes, which is fine because muster does not rely on session continuity for external tokens.
5. Dex rotates refresh tokens on use by default; the intermediary must persist the new one.

**Muster-direct flow** introduces two nested token lifetimes (muster's wrapper + the cached upstream Dex session) and makes pod restarts / horizontal scaling a consistency concern. Not worth it unless muster is the *only* downstream service the intermediary calls.

### Required muster config

Add the intermediary's Dex client ID to `oauth.server.trustedAudiences`:

```yaml
muster:
  oauth:
    server:
      trustedAudiences:
        - <intermediary-dex-client-id>
```

This keeps validation on the **PRIORITY 1** path (local JWKS signature verify — see comment above). Without it, muster falls back to **PRIORITY 2**, which hits Dex's userinfo endpoint on every request: extra latency, tight coupling to Dex availability.

### When muster-as-AS would actually be worth it

Only if:
- The intermediary calls **only** muster and nothing else.
- You need per-client scope carving beyond what Dex offers, enforced at muster issuance time.
- You want to invalidate sessions at the muster layer independently of Dex.

None of that applies to the bedrock-gateway pattern. The intermediary is already the multi-provider policy layer; muster just needs to trust the token it already has.

### TL;DR for the customer

1. Intermediary auths against Dex with `offline_access`.
2. Intermediary holds + rotates Dex refresh token.
3. Inject the Dex `id_token` as bearer to muster (and any other service).
4. Muster config lists the intermediary's Dex client ID in `trustedAudiences`.
5. Refresh proactively before `exp`, not on 401.

PR #576, opened 2026-04-21
fix: forward underlying tool error to call_tool outer isError

### LINKED ISSUE #4281: muster call_tool should forward error
The following example shows a case where `call_tool` is used and an error from the underlying MCP server is returned with isError=false.

Parameters:

```json
{"name":"x_mcp-kubernetes_kubernetes_list","arguments":{"resourceType":"customresourcedefinitions"}}
```

Result:

```json
{
  "content": [
    {
      "type": "text",
      "text": "{\"isError\":true,\"content\":[{\"text\":\"Failed to list resources: failed to list customresourcedefinitions: customresourcedefinitions.apiextensions.k8s.io is forbidden: User \\\"system:serviceaccount:bwi-backstage:mcp-kubernetes\\\" cannot list resource \\\"customresourcedefinitions\\\" in API group \\\"apiextensions.k8s.io\\\" at the cluster scope\",\"type\":\"text\"}]}"
    }
  ],
  "isError": false
}
```

The fact that the top-level `isError` here is `false` is misleading.

Also the question is whether we can avoid the extra response wrapping.

marians:
## Summary

- The `call_tool` meta-tool always returned `isError: false` on the outer response, even when the wrapped tool returned an error. MCP clients inspecting only the top-level `isError` got a misleading success signal.
- Propagate the wrapped tool's `IsError` onto the outer `CallToolResult` in `handleCallTool`, and adjust the BDD test client's `unwrapMetaToolResponse` to keep unwrapping when the outer flag is true (only fall back to "meta-tool error" when content isn't the wrapped JSON shape).

The extra JSON-in-text wrapping is preserved — per the `CRITICAL` comment in `handleCallTool`, BDD tests and clients rely on the wrapped structure. Removing that wrapping is a larger architectural change (threading native `mcp.Content` through `api.CallToolResult` and the aggregator's `convertToMCPResult`) and should be a separate follow-up if desired.

Fixes giantswarm/roadmap#4281

## Test plan

- [x] `go test ./internal/metatools/ -run TestProvider_HandleCallTool -v` — includes a new subtest `forwards underlying tool error to outer IsError`
- [x] `go test ./internal/metatools/ ./internal/testing/...`
- [x] `go build ./...`
- [ ] Manual: call a tool via `call_tool` that errors in the underlying MCP server and confirm the top-level `isError: true` in the response

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 
 - Bump `mcp-oauth` to v0.2.86 with Dex scope filtering: non-standard client scopes like `claudeai` (sent by Claude) are now stripped before forwarding to Dex, preventing `invalid_scope` errors. Also includes Google scope filtering and `openid` force-merge from v0.2.84.
 - CRD validation now uses the discovery API instead of listing `MCPServer` resources in the `default` namespace. With namespace-scoped RBAC (a `Role` limited to muster's own namespace), the previous probe failed with `Forbidden`, silently fell back to filesystem mode, and left configured `MCPServer` CRs unstarted (visible in logs as `Found 0 MCPServer definitions for auto-start processing` followed by `Deleting MCPServer service: <name>`).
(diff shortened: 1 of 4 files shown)


QuentinBisson (approved):
LGTM

PR #591, opened 2026-04-24
Mitigate pre-commit linter complaints

marians:
- Trailing whitespace
- golangci-lint complaints


marians:
No longer needed, main passes pre-commit checks

PR #601, opened 2026-04-29
Introduce spec.transport, remove auth.type=teleport (TB-0/8b)

paurosello:
### What does this PR do?

Lays the **CRD groundwork** for the Muster-on-Gazelle / Teleport Application Access epic ([giantswarm/giantswarm#35456](https://github.com/giantswarm/giantswarm/issues/35456)). Two coupled changes shipped as one release:

**TB-0 (additive)** — new `MCPServer.spec.transport` section:

```yaml
spec:
  transport:
    type: teleport             # discriminator; future values extend additively
    teleport:
      cluster: glean           # the only routing knob
```

The single `cluster` symbol carries no identity. Muster derives both Teleport app names (`mcp-kubernetes-{cluster}`, `dex-{cluster}`) and the matching `tbot-identity-{mcp,tx}-{cluster}` Secret references from it (locked `<role>-<cluster>` convention; consumed by the dispatcher in TB-7 and provisioned by the helm chart in TB-4). Three CRD-level CEL rules guard the field:

1. `transport.teleport` is required when `transport.type=teleport`
2. `transport.teleport` may not be set otherwise (future-proofing the discriminator)
3. `transport` is forbidden when `spec.type=stdio` (no network endpoint)

**TB-8b (removal, breaking but unused)** — drops the legacy `auth.type: teleport` enum value, the `auth.teleport` block, and the `TeleportAuthConfig` Go type. That field was an alpha-only path that bound a single `AppName` to both the MCP HTTP call and the Dex token-exchange call. It is replaced by `spec.transport`, which cleanly separates **transport** (network-level access) from **auth** (identity).

- **Customer Muster (in-VPN):** zero change. `spec.transport` is optional; omitting it preserves direct-HTTPS to `spec.url`.
- **Production CRs:** none affected. `gh search code 'auth.type: teleport' org:giantswarm` confirms no `MCPServer` manifest in any GS GitOps repo uses the removed enum. Phase 5 hasn't shipped.
- **Breaking surface:** any manifest authored against the old shape would now be rejected at admission with `spec.auth.type: Unsupported value: "teleport"`. The new shape lands alongside in TB-7/TB-8 (separate stacked PRs).

### How does it look like?

CEL validation against a live API server (kind, apiserver-side enforcement) — 6/6:

```
[PASS] case-01 transport unset                         (accepted)
[PASS] case-02 transport.teleport.cluster=glean        (accepted)
[PASS] case-03 transport.type=teleport, no .teleport   (rejected: "transport.teleport is required when transport.type is teleport")
[PASS] case-04 transport.type=wireguard                (rejected: "Unsupported value: \"wireguard\": supported values: \"teleport\"")
[PASS] case-05 transport on stdio                      (rejected: "transport is not allowed when type is stdio")
[PASS] case-06 auth.type=teleport (TB-8b)              (rejected: "Unsupported value: \"teleport\": supported values: \"oauth\", \"none\"")
```

- Epic: [giantswarm/giantswarm#35456](https://github.com/giantswarm/giantswarm/issues/35456) — *Muster Access to Private Installations via Teleport Application Access*
- Design review locked the CR-driven dispatcher shape on 2026-04-29. Reverses an earlier "deployment-level routing ConfigMap" sketch in favour of declaring transport intent on the CR.
- Cross-repo follow-ups: per-cluster Teleport app names (`mcp-kubernetes-glean`, `dex-glean`) live in `giantswarm/giantswarm-management-clusters` (TB-1/TB-2). The chart side (`clusters[]` in helm values) lands in TB-4. The runtime dispatcher and aggregator wiring land in TB-7/TB-8. All separate PRs; some stacked on this one.

- Confirm the discriminator shape (`spec.transport.type` enum + `teleport.cluster` single-symbol) is preferred over two free-form fields per CR.
- Confirm the `Removed` CHANGELOG entry adequately flags the breaking-but-unused removal.
- Sanity-check the three CEL rules against the test cases above.

### Testing

- [x] New functionality is covered by unit tests (`pkg/apis/muster/v1alpha1/mcpserver_types_test.go` — 6 round-trip + schema-introspection tests).
- [x] Existing tests have been updated to reflect changes (test cases keyed on `api.AuthTypeTeleport` deleted, not skipped).
- [x] Live CEL validation against a real kube-apiserver — 6/6 cases pass.
- [x] `go build ./...`, `go vet ./...`, `go test ./...` all green.

`docs/reference/crds.md` updated in this PR (drops `teleport` enum row + `TeleportAuth` table). The new `Transport` section — and three audience-targeted how-tos — land in TB-13 (separate PR, stacked).

- [x] CHANGELOG.md has been updated.


paurosello:
## Reshape: `spec.transport.teleport` → explicit fields (PLAN §6 TB-0 revised 2026-04-29)

Pivot point. The previous diff carried a single-symbol `spec.transport.teleport.cluster: <name>` and derived `mcp-kubernetes-{cluster}` / `dex-{cluster}` app names + `tbot-identity-{role}-{cluster}` secret names by convention. Per updated PLAN §9 / §10 / §11, the convention was an *implicit* contract enforced only by docs + helm-test (TB-13's "soft spot"). The shape is now fully explicit:

```yaml
spec:
  transport:
    type: teleport
    teleport:
      mcp:
        appName: mcp-kubernetes-glean
        identitySecretRef: { name: tbot-identity-mcp-glean }
      dex:                                 # optional
        appName: dex-glean
        identitySecretRef: { name: tbot-identity-tx-glean }
```

Adds a fourth CEL rule on top of the three already shipped: `transport.teleport.dex` is required when `spec.auth.type=oauth` AND `spec.auth.tokenExchange.enabled=true`.

**Why:** every Teleport-routed CR is now self-describing and grep-able across the GitOps tree; cluster-admins can name secrets to match their existing tbot-output topology. Trade-off: each CR is ~6 lines longer.

**Migration impact:** none. `gh search` confirms no production CR uses `spec.transport` (the section was added on this same PR and never deployed).

**Live CEL** validated against a kind cluster — 8/8 fixtures behave per PLAN §6 TB-0:

| # | fixture | expected | actual |
|---|---|---|---|
| 1 | transport unset | accept | accept |
| 2 | mcp + dex set, tokenExchange enabled | accept | accept |
| 3 | mcp-only, tokenExchange disabled | accept | accept |
| 4 | type=teleport, no teleport block | reject | reject (CEL) |
| 5 | type=wireguard | reject | reject (enum) |
| 6 | transport on type=stdio | reject | reject (CEL) |
| 7 | dex omitted, tokenExchange.enabled=true | reject | reject (NEW CEL) |
| 8 | mcp.appName violates regex | reject | reject (pattern) |

Cascades to TB-7 dispatcher / TB-8 aggregator / TB-4 helm chart / TB-13 docs follow on local branches.

QuentinBisson:
@paurosello I don't really think having `transport.teleport.mcp` and transport.teleport.dex` is a good idea. 

What do you think about adding a `tranport` for the mcp itself and one for the token exchange?

I would probably split the transport for the mcp under `spec.transport` with possibly a default type of `http` or `public` and have a dedicated `auth` section with a transport under `spec.auth.tokenExchange`. That way, we could have private MCPs behind a public centralized oauth server but also private MCPs behind a private oauth server.



This example is probably a lot more changes that expected and is breaking the CRD API (especially the changes around `auth` to be able to support more than just 1 auth type and oauth provider) but I this is a better direction in the long run:

```
spec:                                                                                                                                                                                                             
    url: https://mcp-kubernetes-finch.teleport.giantswarm.io/mcp                                                                                                                                                    
    transport:                                                                                                                                                                                                      
      type: teleport
      teleport:                                                                                                                                                                                                     
        appName: mcp-kubernetes-finch
        identity:
          secretRef:
            name: tbot-identity-mcp-finch
                                                                                                                                                                                                                    
    auth:
      type: oauth                                                                                                                                                                                                   
      requiredAudiences:
        - dex-k8s-authenticator                                                                                                                                                                                     
   
      tokenExchange:                                                                                                                                                                                                
        enabled: true
        tokenEndpoint:  https://dex-finch.teleport.giantswarm.io/token                                                                                                                                              
        expectedIssuer: https://dex.finch.flex.gigantic.io
        scopes: openid profile email groups                                                                                                                                                                         
        clientCredentialsSecretRef:
          name: muster-tx-credentials-finch                                                                                                                                                                         
                  
        provider: dex
        dex:
          connectorID: gazelle
                                                                                                                                                                                                                    
        transport:
          type: teleport                                                                                                                                                                                            
          teleport:
            appName: dex-finch
            identity:
              secretRef:
                name: tbot-identity-tx-finch
```

paurosello:
@QuentinBisson if we are going in the direction of multiple oauth providers it might be worth it, is this something we are working on?

QuentinBisson:
@paurosello I don't think we are today and we use dex internally no but having the transport at the spec.auth level still makes sense because muster should not be giantswarm specific and this specific way totally is:

```
teleport:
  mcp:
    appName: mcp-kubernetes-glean
    identitySecretRef: { name: tbot-identity-mcp-glean }
  dex:                                 # optional
    appName: dex-glean
    identitySecretRef: { name: tbot-identity-tx-glean }
```

The potential for growth of the second one is a whole lot better and probably not much more work since the Transport struct can be resused

paurosello:
agree, I will do the change

PR #616, opened 2026-05-05
Discovery spec-compliance fixes (WWW-Authenticate, path-based PRM, path-bearing issuers)

QuentinBisson:
Three real spec gaps in muster's per-server OAuth discovery:

- Aggregator's `discoverProtectedResourceMetadata` never parsed `WWW-Authenticate`. MCP 2025-11-25 mandates `resource_metadata=` is consulted first.
- Path-based PRM well-known wasn't probed; only the root form, and it routed through `NormalizeServerURL` which strips `/mcp` and `/sse` from the URL path.
- `pkg/oauth.Client.DiscoverMetadata` only handled no-path issuers. Path-bearing issuers (Microsoft Entra v2.0, Auth0/Okta tenanted) need three URL forms per MCP spec — RFC 8414 path-insert, OIDC path-insert, OIDC append.

Also parses the RFC 9728 `resource` field. Stacked on #615 — review after that lands.


QuentinBisson:

**Why this should merge:** Three real spec gaps in muster's per-server OAuth discovery, all closed:

1. Aggregator's `discoverProtectedResourceMetadata` now parses `WWW-Authenticate: ... resource_metadata=` from a 401 (MCP 2025-11-25 §"PRM Discovery Requirements" mandates this be consulted first).
2. Path-based PRM well-known is now probed before the root form, using the raw MCP URL path so `/v1/mcp` is preserved (`NormalizeServerURL` strips it).
3. `pkg/oauth.Client.DiscoverMetadata` now handles path-bearing issuers (Microsoft Entra v2.0, Auth0/Okta tenanted) — three URL forms per MCP spec: RFC 8414 path-insert, OIDC path-insert, OIDC append.

Plus parses the RFC 9728 `resource` field (logs but accepts when missing — some real backends omit it).

Constants extracted (`WellKnownAuthorizationServer`, `WellKnownOpenIDConfiguration`, `WellKnownProtectedResource`) and reused.

**Self-review pass applied:**
- Fixed botched godoc on `discoverProtectedResourceMetadata` (was missing the function name on first sentence).
- Dropped narrative `issue #599` references from the godoc and Step-0 comment.
- Now strip trailing slash from `parsed.Path` before composing the path-based well-known URL, so `https://x/v1/mcp/` and `https://x/v1/mcp` resolve to the same probe.
- Added a test case for the trailing-slash edge case.

CI green. Stacked on #615 — review after that lands.

paurosello (approved):
Spec-faithful handling of WWW-Authenticate parsing, path-based PRM lookup, and path-bearing issuer normalization, backed by solid test coverage across the discovery edge cases.

PR #617, opened 2026-05-05
Refuse OAuth flow when AS metadata omits S256 PKCE

QuentinBisson:
MCP 2025-11-25 §"Authorization Code Protection" requires clients refuse to proceed when the authorization server's metadata doesn't advertise S256 in `code_challenge_methods_supported`. muster was treating the empty list as "S256 OK" (the OAuth 2.1 default), so both the per-server and agent OAuth flows could start against a server that silently downgrades PKCE.

`Metadata.SupportsPKCE` is renamed to `SupportsS256PKCE`, the empty-list branch returns false, and the check now runs in both `internal/oauth.GenerateAuthURL` (per-server flow) and `internal/agent/oauth.StartAuthFlowWithOptions` (agent flow). Clear error names the issuer.

RFC 8707 (`resource` parameter on auth/token) was originally bundled and is deferred — the full plumbing reaches `state_store` and the token-exchange callbacks. Will file separately if a backend turns out to enforce strict audience binding on the token request.


QuentinBisson:

**Why this should merge:** Closes a real silent-downgrade gap. MCP 2025-11-25 §"Authorization Code Protection" requires clients refuse to proceed when `code_challenge_methods_supported` is absent or doesn't list S256 — muster was treating an empty list as "S256 OK" (the OAuth 2.1 default), which let the AS silently downgrade or reject at the token endpoint with a confusing error.

- `Metadata.SupportsPKCE` renamed to `SupportsS256PKCE` to match the new semantics; only `pkg/oauth`-internal callers existed.
- Refusal is enforced in **both** OAuth flows: `internal/oauth.GenerateAuthURL` (per-server) and `internal/agent/oauth.StartAuthFlowWithOptions` (agent CLI). Error message names the issuer.

**Self-review pass applied:**
- Added `client_pkce_test.go` covering the agent-flow refusal path (the per-server flow already had a test; the agent flow was the asymmetric gap I called out).

CI green. Stacked on #616.

paurosello (approved):
Correctly refuses the OAuth flow when AS metadata is missing, empty, or advertises only plain PKCE — both flows are gated and the error wording makes the failure mode obvious.

PR #619, opened 2026-05-05
refactor(oauth): consolidate JWT payload parsing into pkg/oauth/jwt.go

QuentinBisson:
Four near-identical JWT-payload decoders existed across the tree: `decodeJWTPayload` in the aggregator, `parseIDTokenClaims` in the CLI, `extractIssuerFromToken` in `internal/oauth/token_exchange.go`, and `ExtractEmailFromIDToken` + `decodeSegment` in the admin UI.

Consolidates them into typed accessors in `pkg/oauth/jwt.go` (`Subject`, `Email`, `Expiry`, `Issuer`, `IsExpired` plus an `ErrTokenExpMissing` sentinel). Consumers in `internal/aggregator`, `internal/cli`, and `internal/oauth/token_exchange.go` no longer touch `encoding/base64`, `encoding/json`, or `golang-jwt` directly. The admin diagnostic UI keeps its own segment decoder (lenient display of operator-pasted tokens, including 2-part inputs missing the signature segment).

## Behaviour changes (small, intentional)

- The defensive `RawStdEncoding` fallback for non-spec base64 is dropped from the consolidated path. Every IdP muster integrates with emits RFC 7515-compliant `RawURLEncoding`.
- `internal/admin.DecodeJWT` now accepts 2-part tokens (header.payload) — operators routinely paste truncated tokens from logs.

Both are covered by tests; CHANGELOG entry under `### Changed` documents them.

Issue #549, opened 2026-04-01
SSO token forwarding fails after idle period: stale ID token with zero ExpiresAt never evicted from proxy store


gusevda:
## Summary

After an idle period of >30 minutes with no incoming requests, muster forwards expired JWT ID tokens to downstream `forwardToken: true` MCP servers, resulting in `401 Unauthorized` errors. Token refresh from the client (Backstage) succeeds at the muster level, but the stale ID token continues to be forwarded.

## Root Cause

`storeIDTokenForSSO()` in `internal/aggregator/auth_resource.go:226` stores tokens with only the `IDToken` field populated — no `AccessToken`, no `ExpiresIn`, no `ExpiresAt`:

```go
oh.StoreToken(familyID, userID, musterIssuer, &api.OAuthToken{IDToken: idToken})
```

When `TokenStore.Store()` calls `SetExpiresAtFromExpiresIn()`, it's a no-op (ExpiresIn is 0), leaving `ExpiresAt` as zero. `IsExpiredWithMargin()` in `pkg/oauth/types.go:98` treats zero as "never expires":

```go
if t.ExpiresAt.IsZero() {
    return false // Tokens without expiration don't expire
}
```

This means `GetByIssuer()` returns these stale tokens indefinitely, regardless of the embedded JWT's actual `exp` claim.

## Mechanism

1. User logs in → `SessionCreationHandler` fires → `storeIDTokenForSSO` stores ID-only token with zero ExpiresAt
2. Proactive refresh (triggered by incoming requests) keeps the upstream token fresh and calls `TokenRefreshHandler` → `storeIDTokenForSSO` updates the proxy store
3. After >30 min idle, the upstream Dex token expires and no proactive refresh runs (it only triggers on incoming requests)
4. Next request arrives → client-side refresh succeeds (muster rotates its own tokens), but the upstream Dex token is NOT refreshed → `TokenRefreshHandler` never fires
5. `getIDTokenForForwarding()` reads from the proxy store → gets the stale zero-ExpiresAt token with expired JWT inside
6. Downstream MCP server validates the JWT `exp` claim → rejects with `401 invalid_token`

## Additional Impact

Background SSE listener connections (`listen to server forever`) enter an infinite 1-second retry loop with stale tokens, causing:
- Persistent log flooding: `ERROR: failed to listen to server. retry in 1 second: 401 invalid_token`
- Unnecessary load on downstream servers

### Unit test (instant)

[...]

PR #620, opened 2026-05-05
fix(oauth): set ExpiresAt from JWT exp on stored ID tokens (#549)

QuentinBisson:
Closes part of #549.

`storeIDTokenForSSO` was persisting the proxy-side ID token with zero `ExpiresAt`, so `IsExpiredWithMargin` treated it as never-expiring even after the embedded JWT was past its `exp`. After ~30 minutes idle, downstream servers receive an expired token, reject with 401, and the SSE listeners loop on 1-second retries flooding the logs.

Reads the JWT's `exp` via `pkgoauth.Expiry` (introduced in #619) and persists it as the entry's `ExpiresAt`. JWTs without a parseable `exp` are **refused** at storage time and logged at warn level — the alternative would be to land a zero-`ExpiresAt` entry, recreating the same leak. muster's own ID tokens are minted by mcp-oauth and always carry `exp`, so this only fires on misconfiguration.

Tests cover the three branches: JWT with exp → stored with correct ExpiresAt; unparseable token → refused; JWT missing exp → refused.

This closes the silent-forwarding leg. On-demand upstream refresh (option b in the issue) needs an `mcp-oauth` API addition and will be filed as a follow-up.


QuentinBisson:

**Why this should merge:** Real fix for #549. `storeIDTokenForSSO` was persisting the proxy-side ID token with zero `ExpiresAt`, so `IsExpiredWithMargin` treated it as never-expiring even after the embedded JWT was past `exp`. After ~30 minutes idle, downstream servers receive an expired token, reject with 401, and the SSE listeners loop on 1-second retries flooding the logs.

Reads the JWT's `exp` via `pkgoauth.Expiry` (from #619) and persists it as the entry's `ExpiresAt`. Tokens without a parseable `exp` are **refused** at storage time and logged at warn — landing a zero-`ExpiresAt` entry would recreate the same leak. muster's own ID tokens are minted by mcp-oauth and always carry `exp`, so the refuse path only fires on misconfiguration.

Test trio: JWT with exp → stored with correct ExpiresAt; unparseable token → refused; JWT missing exp → refused.

**Self-review pass applied:**
- Updated PR body to match the code (previously claimed a "graceful fallback" that doesn't exist — the code refuses, the CHANGELOG says refuses, the tests assert refuses).
- Switched to `pkgoauth.Expiry` (replacing the now-removed local `getTokenExpiryTime` from #619). Stack-order resolved.
- Dropped narrative `Issue #549` references from the godoc and test comment.

This closes the silent-forwarding leg. On-demand upstream refresh (option b in the issue) will be a follow-up. CI green.

paurosello (approved):
Tight scoped fix; refuse-to-store-on-parse-error matches issue option (a). Stacked on #619 — must merge that first.

Issue #140, opened 2025-12-19
Subtask 133.7: Architecture Quick Wins


teemow:
## Parent Issue

#133 - Epic: Architecture Improvements for Muster

## Objective

Address small, low-risk improvements that can be completed quickly.

### 1. Add Missing `doc.go` Files

Some packages lack `doc.go` files. Add documentation for:

- [ ] `internal/template/doc.go`
- [ ] `internal/dependency/doc.go` (verify exists)
- [ ] Any other missing packages

### 2. Split Files Exceeding 400 Lines

Per workspace rules, files should be under 400 lines. Current violations:

| File | Lines | Action |
|------|-------|--------|
| `internal/aggregator/server.go` | 1146 | Covered by #133.2 |
| `internal/client/filesystem_client.go` | 1150 | Split by domain |
| `internal/client/kubernetes_client.go` | 500 | Split by domain |
| `internal/workflow/executor.go` | 931 | Extract helpers |

### 3. Remove Deprecated Functions

Remove backward-compatibility aliases:

- [ ] `api.RegisterConfig()` → use `RegisterConfigHandler()`
- [ ] `api.GetConfig()` → use `GetConfigHandler()`

### 4. Consolidate Test Utilities

Move scattered test helpers to `internal/testing/`:

- [ ] Review test files for duplicated mock implementations
- [ ] Consolidate into `internal/testing/mock/`

### 5. Update Import Organization

Ensure consistent import grouping:
1. Standard library
2. External dependencies
3. Internal packages

Run `goimports -w .` across codebase.

## Files to Create

- `internal/template/doc.go`
- Any other missing `doc.go` files

## Files to Modify

- `internal/client/filesystem_client.go` (split)
- `internal/client/kubernetes_client.go` (split)
- `internal/workflow/executor.go` (split)
- `internal/api/config.go` (remove aliases)
- Various test files (consolidate mocks)

## Acceptance Criteria

- [ ] All packages have `doc.go` files
- [ ] No file exceeds 400 lines (except those covered by other issues)
- [ ] Deprecated aliases removed
- [ ] Test utilities consolidated
- [ ] Import organization consistent

## Dependencies

Should be done early - low risk and sets foundation.

## Risks

Low - these are mechanical changes.

## Estimated Effort

Small (1-2 days)

repairman29:
🎯 **PING RECEIVED!**

**@repairman29** has claimed this bounty via [Echeo](https://echeo.io).

**Match Score:** 85%  
[...]

PR #621, opened 2026-05-05
refactor(client/filesystem): move into internal/client/filesystem subpackage (#140)

QuentinBisson:
Closes part of #140.

Moves the 1233-line `internal/client/filesystem_client.go` into a new `internal/client/filesystem/` subpackage, split per CRD domain (`client.go`, `mcpserver.go`, `serviceclass.go`, `workflow.go`, `events.go`). Each file is under 400 lines.

Pure refactor: the `MusterClient` interface stays in the parent `client` package and the only direct caller of `NewFilesystemClient` was the in-package dispatcher, which now calls `filesystem.New(basePath)` directly. External consumers depend only on `MusterClient` and are unaffected.

Renames: `filesystemClient` → `filesystem.Client`, `NewFilesystemClient` → `filesystem.New(basePath string) *Client`, helper writer types (`filesystemStatusWriter`, `filesystemSubResourceClient`) unexported now that the package name disambiguates.

PR #622, opened 2026-05-05
refactor(client/kubernetes): move into internal/client/kubernetes subpackage (#140)

QuentinBisson:
Closes part of #140.

Moves the 547-line `internal/client/kubernetes_client.go` into a new `internal/client/kubernetes/` subpackage, split per CRD domain (`client.go`, `mcpserver.go`, `serviceclass.go`, `workflow.go`, `events.go`). The core file keeps the type, constructor, scheme, lifecycle methods, and discovery-based CRD validation.

Pure refactor: the `MusterClient` interface stays in the parent `client` package and the only direct caller of `NewKubernetesClient` was the in-package dispatcher, which now calls `kubernetes.New(restConfig)` directly. External consumers depend only on `MusterClient` and are unaffected.

Renames: `kubernetesClient` → `kubernetes.Client`, `NewKubernetesClient` → `kubernetes.New(config *rest.Config) (*Client, error)`.


QuentinBisson:

**Why this should merge:** Pure file move. The 547-line `internal/client/kubernetes_client.go` becomes an `internal/client/kubernetes/` subpackage split per CRD (`client.go`, `mcpserver.go`, `serviceclass.go`, `workflow.go`, `events.go`).

- The `MusterClient` interface stays in the parent `client` package; the dispatcher now calls `kubernetes.New(restConfig)` directly.
- Constructor returns a concrete `*Client` (was: `MusterClient`) — embeds `client.Client` so controller-runtime methods stay accessible.
- Constants extracted: `kindMCPServer`, `kindServiceClass`, `kindWorkflow`, `sourceComponent`.
- `validateCRDs` relocated unchanged (discovery-API based — still only checks the MCPServer kind; not regressed, but flagged as an out-of-scope follow-up in #627).

**Per-CRD duplication is intentional in this PR.** Each CRD file is the same template differing only by type and kind string — the symmetric situation to #621. **#627 collapses both client adapters** into shared `client.Object` helpers and normalises error wrapping (workflow.go was previously the inconsistent one).

CI green.

Issue #668, opened 2026-05-13
Replace spec.toolPrefix with spec.instance and unify singleton/family code paths


QuentinBisson:
> **Update 2026-05-19** — original framing (rename-only) was wrong: the rename
> in isolation is a regression in clarity once you look at what
> `toolPrefix` actually does after #670 landed. Rewriting the issue to
> propose the rename **together with** unifying the singleton and family
> code paths.

## Summary

Replace `spec.toolPrefix` on the `MCPServer` CRD with `spec.instance`, and
have the aggregator use `instance` (defaulting to `metadata.name`) as the
family-enum value. Treat singletons as size-1 implicit families internally
so a single code path handles both modes.

## Problem

After #670, `spec.toolPrefix` plays two unrelated roles depending on
whether `spec.family` is set:

| `spec.family` | Effect of `toolPrefix` |
|---|---|
| unset | Drives exposed tool name `x_<toolPrefix>_<tool>` (falls back to `metadata.name`). |
| set   | **Nothing.** `GetAllTools` builds contributions with `serverName: serverName` (`internal/aggregator/registry.go:388`); the family bucket appends `c.serverName` (registry.go:466); `injectInstanceEnum` (registry.go:503) writes those bare `serverName` values into the enum. The family instance identifier callers pass is `metadata.name`, never `toolPrefix`. |

So today `toolPrefix` is half a prefix knob, half a vestigial field, and
the family enum has no override knob at all. Renaming the field to
`instance` without changing behaviour just shifts which half of the
field's job the name lies about.

Concrete bite: the gazelle muster runs 26 kubernetes + 26 prometheus
MCPServer CRs named `<cluster>-mcp-<type>` for kustomize / label hygiene.
Callers have to pass `management_cluster=agama-mcp-kubernetes` because
the family enum is hard-wired to `metadata.name`. They want to pass
`management_cluster=agama`. There is no field that lets them.

## Proposal

1. **Add `spec.instance`** — `^[a-zA-Z][a-zA-Z0-9_-]*$`, optional,
   defaulting to `metadata.name`.
2. **Aggregator changes** (`internal/aggregator/registry.go`):
   - Non-family servers: exposed tool name = `x_<instance>_<tool>`
     (today's behaviour, with `instance` substituted for `toolPrefix`).
   - Family servers: `injectInstanceEnum` writes `instance` values
     (not raw `serverName`). `family.name` still drives the tool-name
     prefix.
[...]

PR #670, opened 2026-05-13
Group equivalent MCP servers under spec.family (supersedes #543)

QuentinBisson:
## Summary

Supersedes #543 (Théo's auto-dedup approach). Addresses @teemow's [CHANGES_REQUESTED review](https://github.com/giantswarm/muster/pull/543#pullrequestreview-vICMCO_LZqc7), the four inline bugs flagged in [Quentin's review](https://github.com/giantswarm/muster/pull/543#pullrequestreview-77u2FH_LZqc77u2FH), and [teemow's follow-up feedback on this PR](https://github.com/giantswarm/muster/pull/670#issuecomment-4447540998).

Adds `MCPServer.spec.family` as an explicit, opt-in declaration that an instance is one of a family of equivalent servers (e.g. multiple kubernetes MCP servers pointed at different clusters). The aggregator exposes tools from family members under `x_<family.name>_<tool>` with a required parameter (named by `family.instanceArg`) that selects the backend. Non-family servers keep today's per-server prefixing — no behaviour change for existing CRs.

## Why explicit, not automatic

teemow's review on #543 flagged the auto-dedup model as surprising and unstable: a single-server fleet got per-server names today, then silently switched to family-style names the moment a second instance joined. `spec.family` makes the contract explicit, declared by the operator. The instance-arg is required even for single-instance families so skills written against the family name continue to work as the fleet grows.

## Spec shape

```yaml
spec:
  family:
    name: kubernetes
    instanceArg: management_cluster
```

Both fields required when `family` is set — structural validation handles the linkage, no CEL needed. `instanceArg` is per-family configurable so the LLM-facing surface reads naturally for the domain (`management_cluster`, `country`, `model`, …) instead of a generic `server`.

## Design highlights

| Scenario | Exposed |
|---|---|
| One server, `family: {name: kubernetes, instanceArg: management_cluster}` | `x_kubernetes_<tool>` with required `management_cluster` enum (single value) |
| Two servers, same family | `x_kubernetes_<tool>` with required `management_cluster` enum listing both instances |
| No `family` | `x_<toolPrefix-or-name>_<tool>` (today's behaviour) |
| Same family, diverging tool description | That specific tool falls back to per-server prefixing (warn); other tools still group |
| Same family.name, diverging family.instanceArg | Entire family falls back to per-server prefixing (warn) — enforced at the family level (Pass 1 in `assembleExposedTools`), so disjoint tool sets cannot escape detection |

`toolPrefix` is kept untouched (no breaking change). Once `spec.family` is widely adopted, the post-family role of `toolPrefix` is purely an instance label — see #668 for the follow-up rename proposal.

## Bug fixes addressed from #543's review

- **Dedup key on `(name, description)`, not just name** — diverging descriptions trigger per-server fallback for that tool, with a warning log. Preserves backends that ship older/different versions of the same tool.
- **Deep-copy `InputSchema.Properties` / `Required`** — `injectInstanceEnum` builds fresh map and slice so injecting the instance-arg enum does not mutate the backend's cached tool. Regression test enforces this across repeated `GetAllTools` calls.
- **`Deregister` cleans reverse-lookup state** — walks `nameMapping` and `familyMappings` and drops entries owned by the deregistering server. Family entries with surviving providers keep them; empty entries are removed entirely.
- **Explicit-but-invalid instance arg surfaces the resolution error** — `ResolveToolNameForServer` returns "tool X is not available on server Y (available: ...)" and `CallToolInternal` returns it directly instead of silently falling back to legacy `nameMapping` resolution.
- **No prefix-coupled family lookup** — family routing stores the family name + instance-arg alongside providers, so `ResolveToolName` reads the family directly off the bucket instead of trimming `musterPrefix+"_"` from the exposed name.

## Bugs surfaced by 367531a and fixed in 5ae7317

Two reproducers added in commit 367531a documented behavioural gaps in the initial implementation. Both are now resolved:

- **Cross-session corruption** (`mcpserver-family-cross-session-corruption`) — `GetAllToolsForSession` ran `assembleExposedTools`, which reset and rewrote the process-global `familyMappings` index based on the calling session's session-scoped contributions. A second user listing tools therefore overwrote the first user's view of family membership, and subsequent `ResolveToolNameForServer` calls rejected legitimate explicit-instance routing.
  - Fix: replace the reset-and-rewrite with `upsertFamilyTool`, which unions providers into existing entries. `Deregister` remains the only path that removes providers from the index, so cleanup semantics are unchanged.
- **Disjoint-tools divergence** (`mcpserver-family-disjoint-tools-divergence`) — divergence detection was per-`(family, tool)` bucket and only fired when a second contributor joined the same bucket. Family members with disjoint tool sets each occupied their own bucket, so a divergent `family.instanceArg` was never seen, and the aggregator emitted multiple `x_<family>_<tool>` entries with conflicting instance-selector args.
  - Fix: promote `instanceArg` validation to a family-level invariant driven by the global `serverFamilies` map. `familyFallbackStatusLocked` computes the divergent-family set independently of tool overlap. Members of a fallback family emit per-server prefixed names for ALL their tools. `Register` / `RegisterPendingAuth` purge any stale family-grouped entries the moment a divergent member joins.

## API change (no breaking config)

`aggregator.Register` / `aggregator.RegisterPendingAuth` and their manager-level / `api.AggregatorHandler` counterparts now take a `ServerRegistration` / `PendingAuthRegistration` struct rather than five-to-six same-typed positional string args. `RegisterPendingAuth` / `RegisterPendingAuthWithConfig` collapse into a single method — `AuthConfig` is a nullable field on the struct. Internal-only change.

## Tests

- `TestServerRegistry_FamilyGrouping` — eight subtests covering single-instance, multi-instance, mixed family+non-family, description divergence, `Deregister` cleanup, missing-`instanceArg`-arg error, unknown-`instanceArg`-arg error, and the deep-copy regression.
- `TestAggregatorServer_CallToolInternal_FamilyRouting` — three subtests covering explicit-instance routing, invalid-`instanceArg` error surfacing, and the missing-`instanceArg` error path.
- Behavioural scenarios under `internal/testing/scenarios/mcpserver-family-*.yaml` exercising the full muster-serve loop:
  - `mcpserver-family-single-instance`, `multi-instance`, `mixed` — happy-path coverage
  - `mcpserver-family-description-divergence`, `instance-arg-divergence` — fallback paths
  - `mcpserver-family-deregister` — two-member deregister cleanup
  - `mcpserver-family-cross-session-corruption`, `disjoint-tools-divergence` — bug reproducers (now passing)
  - `mcpserver-family-three-member-partial-deregister` — broader deregister coverage on a three-instance family with the middle member removed, asserting the survivors stay routable and the rejected-instance error names both survivors
  - `mcpserver-family-partial-description-divergence-strict` — pins the negative assertion that the diverging tool's family-prefixed name is absent (the existing description-divergence scenario only checked the per-server fallback works and the matching tool still groups)
- All existing tests pass; full `muster test --concept mcpserver` suite (67 scenarios) is green; `make lint` clean.

## Follow-ups

Tracked in #697:
- Deep-copy of nested input-schema properties (currently only top-level `Properties`/`Required` are cloned).
- Deterministic ordering in `tools/list`.
- `family.instanceArg` collision with a tool-defined input property (three viable fix paths — needs maintainer decision).
- Optional reconciler-level fail-fast on `family.instanceArg` conflict (depends on extending `MCPServerInfo` to expose `Status.Conditions[]`).

Out of scope for this PR:
- #668 — rename `toolPrefix` once `family` is the canonical naming source.
- A separate `MCPServerFamily` CRD (family as a first-class resource with its own description / RBAC / cross-cluster discovery) is the cleaner long-term shape if the family surface grows beyond name + instanceArg; tracked as a future issue.

Closes #543 by replacement.


teemow:
@QuentinBisson I haven't tested it, but here are a few things that you should be careful with:

* The family pattern needs to be independent of the muster prefix. If a user changes the muster prefix from x_ to y_ the family configuration should not be affected.
* Imo it would be good to define a parameter name for the server too. This would make it semantically easier for the LLM. So in our case the server parameter would be called management_cluster. But in other cases it can also be countries or cars.
* Please add behavioral scenarios that test this thoroughly. This ensures that this works correctly and that this functionality isn't destroyed accidentally

QuentinBisson:
Thanks @teemow, all three land.

On the prefix independence, `spec.family` is already stored separately from `musterPrefix` and the family routing index gets rebuilt on every tool-list assembly, so a prefix change can't leave stale entries. There's one spot in `ResolveToolName` where I derive the family name back out of the exposed name (just for the error string) and I'll fix that to read it off the provider record directly. Nothing else couples the two.

On the parameter name, yeah, you're right, `server` is too generic for the LLM to do anything useful with. I went back and forth on whether to add a sibling field or fold everything under `family`, and I think nesting is cleaner:

```yaml
spec:
  family:
    name: kubernetes
    instanceArg: management_cluster
```

Both required. Then the structural validation handles "you can't set the arg without naming the family" for free, and there's room to add more family-level config later (description, etc.) without polluting the top of the spec. The aggregator picks up `instanceArg` per family and injects it instead of the hard-coded `"server"`. The PR isn't merged yet so the `family: <string>` to struct change has zero migration cost.

On behavioral scenarios, fair, the Go unit tests aren't enough. I'll add scenarios under `internal/testing/scenarios/mcpserver-family-*.yaml` covering single-instance, multi-instance, mixed family + non-family, description divergence, deregister cleanup, and the missing/unknown `instanceArg` paths. Basically the eight subtests in `TestServerRegistry_FamilyGrouping` transcribed into YAML so they exercise the full muster-serve loop, not just the registry.

Will push the nested struct + scenarios as separate commits.

QuentinBisson:
@teemow we're good :)

teemow (approved):
Approving. The fix correctly addresses both review findings:

- **Cross-session corruption**: `upsertFamilyTool` unions providers into the existing bucket; `Deregister` remains the sole removal path. A session's tools/list can no longer erase another session's view.
- **Disjoint-tools divergence**: `familyFallbackStatusLocked` lifts `instanceArg` divergence detection to the family level driven by the global `serverFamilies` map, independent of which tools any pair of members happens to share. Members of a fallback family emit per-server prefixed names for all their tools, with a single state-change warning at registration.

Verified locally: both reproducers pass; full `muster test --concept=mcpserver` suite (67 scenarios) green; `internal/aggregator/...` unit tests green. The two additional scenarios (`three-member-partial-deregister`, `partial-description-divergence-strict`) harden adjacent paths the refactor touches.

PR #684, opened 2026-05-18
refactor(aggregator): adopt mcp-go native OTEL tracing hooks

QuentinBisson:
## Summary

Replace muster's per-tool-handler OTEL tracing middleware with mcp-go's native server-tracing hooks via the [`mcp-go/otel`](https://github.com/mark3labs/mcp-go/tree/v0.54.0/otel) adapter (the feature I contributed upstream in [mark3labs/mcp-go#856](https://github.com/mark3labs/mcp-go/pull/856), v0.54.0).

The custom middleware only wrapped `tools/call`. The native hooks cover every dispatched JSON-RPC method (`initialize`, `tools/list`, `resources/read`, `prompts/get`, …) plus the tool-handler span, and extract W3C trace-context from inbound headers.

## What changed

| | Before | After |
|---|---|---|
| JSON-RPC method spans | none | `mcp.<method>` (SpanKindServer) on every dispatch |
| Tool-handler spans | `tool.<name>` from muster's middleware | `tool.<name>` from mcp-go's middleware (same scope, same attrs) |
| Trace propagation | none on the server side | W3C `traceparent` extracted from inbound HTTP headers |
| Custom code | `instrument.Tracing()` middleware + test (~85 lines) | gone |

`instrument.StartToolSpan` is retained — it's still used by `CallToolInternal` to capture the backend-resolved tool name for the internal dispatch path (workflows, direct API), which doesn't go through mcp-go's handler middleware.

The OTEL option chain (`mcpotel.WithServerTracing` + Logging middleware + Metrics middleware) now lives in `instrument.MCPServerOptions`, called from `server.Start`. The new `TestMCPServerOptions_HistogramExemplarAttachesToolHandlerSpan` builds an `MCPServer` with that helper, dispatches a `tools/call` through `HandleMessage`, and asserts the duration histogram observation carries an exemplar with the `tool.<name>` span's `TraceID`. If a future change drops or no-ops `WithServerTracing` the test fails — pinning the contract reviewer called out.

## Stack

The OAuth test-constant lift (`chore(oauth): hoist repeated test literals to package constants`) has moved out of this branch into its own prep PR so the OTEL diff is review-only OTEL. This branch is now `OTEL refactor + wiring test`; the OAuth lift PR (linked separately) should land before this one and before #685.

## Validation

- `go test ./...` clean
- `make lint` clean (no new issues; pre-existing goconst/gosec/govet warnings in other packages are unchanged)
- 121/121 behavioural scenarios pass via `./muster test --parallel 50 --base-port 30000`


QuentinBisson (commented):

The "adopt mcp-go native, drop hand-rolled" goal is met — minimum muster-specific glue retained.

**Non-blocking**
- The OAuth test-constant churn in `internal/oauth/*` and `internal/agent/oauth/token_store_test.go` is unrelated to OTEL and stacked through #684/#685/#696. Lifting it to a prep PR would roughly halve the review surface for each tracing PR.

**Tests**
- `TestTracing` correctly deleted (code gone). Nothing new asserts that `mcpotel.WithServerTracing(...)` is wired outermost in `server.Start`. An option-order regression would slip past current tests. One integration-style test exercising `NewMCPServer` with the production middleware chain and asserting an exemplar would be cheap insurance.

paurosello:
Resolved CHANGELOG conflict against main (`283b6f8`, merge — kept both entries).

**Merge order: #684 → #685 → #696.** This PR lands first: installs server-side `mcpotel.WithServerTracing` (inbound half). #685 then adds outbound client tracing on top. #696 last — it deletes `instrument.StartToolSpan` that this PR still relies on for `CallToolInternal`.

OAuth test-constant lift is still worth doing (see thread above) but not gating approval.

paurosello (approved):
Approving. OTEL change is clean; per-handler middleware → native mcp-go hooks is the right direction.

PR #685, opened 2026-05-18
feat(mcpserver): trace outbound MCP client calls via mcp-go/otel

QuentinBisson:
## Summary

Install mcp-go's OTEL tracer + W3C trace-context propagator on every outbound mcp-go client (stdio, SSE, streamable-http, dynamic-auth) by applying [`mcp-go/otel.WithClientTracing`](https://github.com/mark3labs/mcp-go/blob/v0.54.0/otel/options.go) as a post-construction `ClientOption` on the returned `*client.Client`.

The muster → backend leg now:

- inherits the trace context from the inbound caller's request
- emits a W3C `traceparent` header on every outgoing JSON-RPC frame
- produces `mcp.<method>` spans of kind Client at the muster side

Pairs with #684 (server-side tracing, merged) so a single trace covers the whole caller → muster → backend chain.

## Implementation note

mcp-go's convenience client constructors (`NewStdioMCPClient`, `NewSSEMCPClient`, `NewStreamableHttpClient`) don't accept `client.ClientOption` — only their respective `transport.*Option`. So `WithClientTracing` is applied post-construction by invoking the option directly on the returned `*client.Client`. This is safe because `ClientOption` is `func(*Client)` field-assignment and the tracer is read only at request time.

Upstream improvement worth pursuing: the convenience constructors should accept `...client.ClientOption` so users don't need to know about the two-layer transport/client option split.

## Review follow-ups

- `TestClientTracing_PostConstruction` pins the contract: builds an in-process MCP server, constructs a real `*client.Client`, applies `mcpotel.WithClientTracing(otel.Tracer(instrument.TracerName))` post-construction (exact production pattern), dispatches `tools/call`, and asserts an `mcp.tools/call` span of kind `Client` is recorded. Catches an upstream `ClientOption` signature change and any regression where the option-application line is removed at a call site.
- Tracer scope folded: `instrument.TracerName` renamed from `github.com/giantswarm/muster/internal/aggregator` to the module path `github.com/giantswarm/muster`. Both legs (inbound server, outbound client, `StartToolSpan`) now emit under one scope — direction is already in `SpanKind` + span name.
- No `applyClientTracing` helper: the option-application is a one-liner, inlined at the four call sites instead of being wrapped in a trivial passthrough.
- OAuth test-constant lift moved out into a standalone prep PR.

## Validation

- `go test ./...` clean
- `make lint` clean
- 121/121 behavioural scenarios pass via `./muster test --parallel 50 --base-port 30000`


QuentinBisson (commented):

**Non-blocking**
- `applyClientTracing` invokes `mcpotel.WithClientTracing(...)(c)` directly because the convenience constructors don't accept `client.ClientOption`. You called this out in the PR body — fine — but it's a latent fragility. If mcp-go ever changes `ClientOption` to take a `context` or return an error, this silently breaks. Add a unit test that constructs a real `*client.Client`, calls `applyClientTracing`, makes one outbound call, and asserts a client-kind span. That pins the contract.
- YAGNI: `if c == nil { return }` is defensive against a path no caller reaches (every caller checks the constructor error and bails). Drop the guard or document the one nil path.

**Tests**
- No tests for the new `tracing.go` file.

Same OAuth test-constant churn as #684 bleeds in via the stack.

paurosello:
Resolved CHANGELOG conflict against main (`a2f8972`, merge — kept both entries). On the merging order:

**#684 → #685 → #696.** Reasoning:
1. **#684 first** — installs server-side `mcpotel.WithServerTracing` and is the inbound half of the trace chain. Outbound (#685) without inbound = orphan client spans with no parent context.
2. **#685 next** — adds outbound client tracing on top of #684. Will be a clean fast-forward once #684 lands and the OAuth test-constant collision is resolved (see #684 thread — folding into #706 keeps the OTEL PRs lean).
3. **#696 last** — deletes `instrument.StartToolSpan` that #684 still relies on for `CallToolInternal`. Landing it before #684 would break the build.

Approving as the OTEL change itself is clean.

paurosello (approved):
Approving the OTEL change. Merge order applies (see comment above) — block on #684 landing first.

PR #687, opened 2026-05-18
Remove Teleport authentication support

QuentinBisson:
## Summary

Teleport Application Access integration is moving out of muster into a dedicated operator, so this PR rips the whole thing out of the main codebase.

## What's gone

- **`internal/teleport` package** — cert watcher, mTLS HTTP client, security helpers, adapter, tests (about 3.5k lines).
- **API surface (`internal/api`)** — `TeleportClientHandler` interface, `RegisterTeleportClient` / `GetTeleportClient`, `TeleportClientConfig`, `TeleportAuth`, `AuthTypeTeleport`, and the entire `internal/api/teleport.go` file.
- **MCPServer CRD (`muster.giantswarm.io/v1alpha1`)** — `TeleportAuthConfig` type and the `MCPServerAuth.Teleport` field are deleted. The `auth.type` enum is narrowed from `oauth;teleport;none` to `oauth;none`. Helm CRD and deepcopy are regenerated.
- **OAuth & token exchange** — `OAuthHandler.ExchangeTokenForRemoteClusterWithClient`, `TokenExchanger.ExchangeWithClient`, and their adapters. The only consumer was the Teleport mTLS path; the standard `ExchangeTokenForRemoteCluster` covers everything else.
- **Streamable-HTTP client plumbing** — `MCPClientConfig.HTTPClient`, `NewStreamableHTTPClientWithHTTPClient`, `NewStreamableHTTPClientWithHeaderFuncAndHTTPClient`. Same story: only fed by the Teleport path.
- **Aggregator helper** — `getTeleportHTTPClientIfConfigured` and the `TeleportClientResult` dual-path branches in `connection_helper.go` and `server.go`. Token-exchange and MCP-client construction are now single-path.
- **BDD scenarios** — the four `mcpserver-teleport-*.yaml` scenarios; the proxied token-exchange scenario keeps its proxy semantics but no longer name-drops Teleport.
- **Docs** — `docs/how-to/teleport-authentication.md` deleted; README, architecture, capabilities, problem-statement, configuration-examples, getting-started, mcp-tools, and CRD reference scrubbed of every Teleport reference. Example tool names that used `x_teleport_*` now use `x_kubernetes_*` / `x_prometheus_*`.
- **Helm** — RBAC and values comments no longer mention teleport identity secrets.

## Breaking change

Existing MCPServer CRs with `auth.type: teleport` or any `auth.teleport` block will be rejected by CRD validation after upgrade. Migrate those workloads to the new operator before rolling this out.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` — clean
- `make helm-lint`, `make helm-test` — clean
- `muster test --generate-schema` regenerates `schema.json` with no teleport entries
- Final `grep -ri "teleport\|tbot"` returns only the CHANGELOG entry that documents the removal

Net diff: **57 files, 59 insertions, 6091 deletions**.


QuentinBisson (commented):

Removal is clean. Spot-checked: `schema.json` enum narrowed and `teleport` map removed; `mockTeleportClientHandler` gone; `mockOAuthHandler.ExchangeTokenForRemoteClusterWithClient` stub gone; helm RBAC comments scrubbed; `TestConfigurationChanged` case rewritten to `Type: "none"` rather than just deleted; whole BDD scenarios deleted as a batch (right call — they were Teleport-specific). Diff `+`-side grep for `teleport|tsh|tbot` returns only the CHANGELOG line.

**Non-blocking — YAGNI**
- `validateExchangeRequest` and `getExchangeDefaults` (`internal/oauth/token_exchange.go`) were the DRY seam between `Exchange` and `ExchangeWithClient`. With `ExchangeWithClient` gone they each have a single caller. Inline back into `Exchange` or keep — defensible either way; flagging only because it was the seam's only reason to exist.

paurosello (approved):
Verified hard removal is clean:

- `internal/teleport/` deleted as a unit (package + API surface)
- `auth.type` CRD enum narrowed to `oauth;none` (`helm/muster/crds/muster.giantswarm.io_mcpservers.yaml`)
- `schema.json` has zero `teleport` matches
- `pkg/apis/muster/v1alpha1/mcpserver_types.go` — no `TeleportAuthConfig` type
- BDD scenarios scrubbed; `TestConfigurationChanged` rewritten sensibly to `Type: "none"`
- `go build ./...` and `go vet ./...` clean

External-consumer check (giantswarm org-wide GH code search):
- 0 hits for `auth.type: teleport` in any MCPServer manifest
- 0 hits for `TeleportAuthConfig`, `AuthTypeTeleport`, `muster/internal/teleport` outside this PR

CHANGELOG entry is migration-aware. No backwards-compat shim — clean break, as intended.

PR #688, opened 2026-05-18
feat(agentgateway): PR 1/16 — pure agentgateway domain

QuentinBisson:
Step **1/9** of the agentgateway adapter series. Plan: `/home/quentin/.claude/plans/declarative-forging-lynx.md`.

## What this lands

Pure Go agentgateway domain — no Kubernetes client, no filesystem I/O, no agw runtime.

- `internal/reconciler/agentgateway/doc.go`
- `internal/reconciler/agentgateway/config.go` — `Config{Name, Namespace, Backends, Routes, Policies}`, `Backend{Name, Target}`, `Target` interface with `Kind() TargetKind`, `HTTPTarget{Protocol, Host, Port, Path}`, `StdioTarget{Command, Args, Env}`, `Route`, `Policy`, `Authn`, `TokenExchange`, `AuthorizationServer`, `AuthnType`
- `internal/reconciler/agentgateway/new_config.go` — `NewConfig(name, namespace string, spec) (Config, error)`
- `internal/reconciler/agentgateway/new_config_test.go` — table-driven happy and error cases
- `internal/reconciler/agentgateway/fixtures_test.go` — auth profile / spec builder fixtures

## Why the package + names changed

`translator/` + a separate `emitter/` was an artificial split — nothing in the names said "agentgateway configuration". Lean into the actual domain:

- Package `translator` → `agentgateway`
- Type `Model` → `Config` (what it actually is)
- Function `Transform`/`Compile` → `NewConfig` — idiomatic Go constructor; stdlib precedent `http.NewRequest`, `template.New`, `json.NewDecoder`

At call sites:

```go
config, err := agentgateway.NewConfig(name, namespace, spec)
if err := k8s.NewApplier(client, ownerRef, cfg).Apply(ctx, config); err != nil { ... }
```

## Backend as a sum type, not a tagged-pointer struct

`Backend{Name, StreamableHTTP *HTTPTarget, SSE *HTTPTarget, Stdio *StdioTarget}` enforced "exactly one" at runtime via three pointer nil-checks. Replaced with a compile-time sum type:

```go
type Backend struct {
    Name   string
    Target Target
}

type Target interface { Kind() TargetKind }

type HTTPTarget struct { Protocol HTTPProtocol; Host string; Port int; Path string }
func (HTTPTarget) Kind() TargetKind { return TargetHTTP }

type StdioTarget struct { Command string; Args []string; Env map[string]string }
func (StdioTarget) Kind() TargetKind { return TargetStdio }
```

`StreamableHTTP` and `SSE` collapse into one `HTTPTarget` carrying a `Protocol` field — both transports already share host/port/path. Adapters type-switch on `b.Target.(type)`.

## DDD/hexagonal split on identity

`Config` carries `Name` + `Namespace` — the MCPServer identity both adapters need. K8s-specific persistence fields (`UID`, `APIVersion`, `Kind` for `ownerReferences`) stay out of the domain; they bind on the K8s adapter's constructor (`k8s.NewApplier(client, ownerRef, cfg)`, per-Reconcile). The YAML adapter never sees them.

## Teleport authentication removed

`AuthnTypeTeleport`, `TeleportAuthn`, the `auth.Teleport` branch in `NewConfig`, `parseAuthnType` acceptance of `"teleport"`, and every teleport test case are gone from this package. `NewConfig` rejects `spec.auth.type == "teleport"` as an unsupported type. The `MCPServerAuth.Teleport` CRD field is removed in a separate change.

## Behaviour

- `spec.type == "stdio"` → `Backend{Target: StdioTarget{Command, Args, Env}}`.
- `spec.type ∈ {"streamable-http","sse"}` → parse `spec.url`, derive `Host`, `Port` (default `80` for `http`, `443` for `https`), `Path`. Populate `HTTPTarget` with the matching `Protocol`.
- `spec.auth` derives an `Authn` capturing `type`, `forwardToken`, `requiredAudiences`, `tokenExchange`, and `authorizationServer`. CRD mutual-exclusion rules are enforced.
- `NewConfig` is pure: same input → same output, no I/O.

## Tests

- `TestNewConfig_Happy` — **21** table cases (3 `spec.type` × 7 auth profiles)
- `TestNewConfig_Errors` — **16** cases (missing name, missing namespace, missing fields, invalid URLs, CRD mutex violations, unknown enums, teleport rejection)
- `TestNewConfig_IsPure` — purity regression
- `TestNewConfig_HTTPStreamableWithExplicitPort` / `TestNewConfig_HTTPDefaultPort` — URL parsing edges
- `TestNewConfig_StdioPreservesArgsAndEnv` — stdio target carries args/env verbatim

`go test -race ./internal/reconciler/agentgateway/...` clean; **100% statement coverage**. `golangci-lint run` clean.


QuentinBisson (commented):

**DDD**
- Domain package imports only `fmt`, `net/url`, `strconv`, and `pkg/apis/muster/v1alpha1`. The v1alpha1 dependency leaks in via `NewConfig(spec MCPServerSpec)`. `NewConfig` is really a translation function (CRD shape → pure domain values) — consider extracting it to `agentgateway/translate` (or moving to the reconciler that owns the CRD watch) so the domain package itself has zero CRD types. Acceptable trade-off today; flag if a second producer ever needs `NewConfig`.

**Non-blocking**
- `httpTargetFromURL` silently swallows `strconv.Atoi` error via `port, _ = strconv.Atoi(p)`. `url.URL.Port()` enforces digits, so it's safe in practice — but the test `streamable-non-numeric-port` only passes because `url.Parse("http://host:abc/mcp")` itself errors. Drop the `_` and return explicitly.
- `parseAuthnType` accepts `""` as `AuthnTypeNone` while non-empty `AuthorizationServer` with `Type == ""` is rejected. Correct, but document the "can't omit type and still pin an authorization server" trap in CRD validation or PR body.
- `Authn.RequiredAudiences` is consumer-trusted (no dedup, no order normalization, no emptiness check). Call it out in the godoc.

**Suggestion**
- `TestNewConfig_IsPure` tests *determinism*, not purity (no I/O is observed). Rename to `TestNewConfig_IsDeterministic`. If you want real purity coverage, run under `-count=2` with a poisoned env, but the rename is what the name overpromises today.

**Tests**
- 21 happy × 16 error cases is solid. Missing: SSE with explicit port; stdio with nil args/env; `RequiredAudiences` dedup/order if you ever care.

PR #689, opened 2026-05-18
feat(agentgateway): PR 2/16 — Applier interface

QuentinBisson:
Step **2/9** of the agentgateway adapter series. Plan: `/home/quentin/.claude/plans/declarative-forging-lynx.md`.

## What this lands

- `internal/reconciler/agentgateway/applier.go` — two adapter-agnostic port interfaces:

```go
type Applier interface {
    Apply(ctx context.Context, config Config) error
}

type Deleter interface {
    Delete(ctx context.Context, name string) error
}
```

No tests in this PR — the contract is exercised by PR 3 (K8s applier) and PR 4 (YAML applier) against actual implementations. The merge of `Applier` and `Deleter` into one port stays in PR 12 where the rest of the reconciler-shape collapse lives (`Service` wrapper drop + single reconciler constructor).

## Contract

Implementations must be idempotent, reconcile prior state for entities they own, honour context cancellation, and wrap returned errors so callers can `errors.Is` / `errors.As`. `Deleter` of an absent name returns `nil`. `Deleter` is held separately because not every `Applier` needs an explicit delete; callers that don't need it pass `nil`.

## Hexagonal shape

`Config` is the domain port. The two interfaces stay free of K8s and filesystem vocabulary — ownerRef, dir, client are bound on the adapter's constructor (`k8s.NewApplier(client, ownerRef, cfg)` per Reconcile in cluster mode; `yaml.NewApplier(dir)` long-lived in filesystem mode). The port doesn't reach into adapter internals in either direction.

## Notes

- Built on PR #688's `agentgateway` package.
- `Apply` takes `Config` (the renamed `Model`).


QuentinBisson (commented):

ISP and DIP honored — one-method port on the domain side, both adapters depend inward.

**Non-blocking — contract wording**
- Doc prescribes errors wrap as `fmt.Errorf("context: %w", err)` — that's a literal-format requirement. The K8s applier in #692 uses `"k8s applier: backend %q: %w"` and would violate its own contract on paper. What the contract actually wants is "errors must wrap so `errors.Is`/`errors.As` traverse the cause". Loosen the doc.
- Doc says "reconcile prior state: when an Apply drops entities previously persisted, the persisted view is updated." Today the K8s applier honors this only for the policy (drop-forward-token deletes it); it does NOT delete backends or routes that disappear from the Config. Either tighten the contract ("single-MCPServer single-backend") or expand the impl (list-and-prune). Theoretical today since Config has one backend per MCPServer, but the contract should be honest.

**Suggestion**
- The yaml applier exposes `Delete` publicly; the port doesn't. If `Delete` is part of the lifecycle the reconciler drives, hoist it onto the port. If cluster mode genuinely relies only on ownerRef cascade, document the asymmetry on `applier.go` so reviewers don't grep for the symmetry that isn't there.

**Tests**
- No tests in this PR is fine (the contract is exercised by impl tests in #692/#695). Mention this in `applier.go`'s doc so reviewers don't grep for a missing `applier_test.go`.

paurosello (approved):
LGTM. Small, focused Applier port. Domain stays clean.

PR #692, opened 2026-05-18
feat(agentgateway): PR 3/16 — K8s emitter (cluster mode)

QuentinBisson:
Step **3/9** of the agentgateway adapter series. Plan: `/home/quentin/.claude/plans/declarative-forging-lynx.md`.

## What this lands

The K8s `Applier` — implementation of `agentgateway.Applier` (port from PR #689) for cluster mode.

```go
applier := k8s.NewApplier(client, ownerRef, k8s.Config{
    GatewayName:      "muster-agw",
    GatewayNamespace: "muster",
})
applier.Apply(ctx, config)
```

`NewApplier` is constructed per Reconcile, bound to the MCPServer's `metav1.OwnerReference` so the emitted stack cascade-deletes.

Files:
- `internal/reconciler/agentgateway/k8s/doc.go`
- `internal/reconciler/agentgateway/k8s/applier.go` — `Applier`, `NewApplier`, `Apply`, `ErrStdioNotSupportedInCluster`, `Config`
- `internal/reconciler/agentgateway/k8s/reconcile.go` — per-object reconciliation (`AgentgatewayBackend`, `HTTPRoute`, `AgentgatewayPolicy`)
- `internal/reconciler/agentgateway/k8s/helpers.go` — `createOrUpdate` with conflict retry, `deleteIfExists`, `applyOwner`, protocol mapping, policy spec derivation
- `internal/reconciler/agentgateway/k8s/*_test.go` — full coverage (idempotency, stdio rejection, drop-forward-token cleanup, owner-ref validation, context cancellation, conflict retry, persistent-conflict failure, cascade owner refs)

## Upstream CRD types imported, not redeclared

Removed `internal/agentgateway/apis/v1alpha1/` (hand-rolled `AgentgatewayBackend` + `AgentgatewayPolicy` structs). The Applier imports the real types directly:

- `github.com/agentgateway/agentgateway/controller/api/v1alpha1/agentgateway` — `AgentgatewayBackend`, `MCPBackend`, `McpTargetSelector`, `McpTarget`, `MCPProtocol`, `AgentgatewayPolicy`, `BackendFull`, `BackendSimple`, `BackendAuth`, `BackendAuthPassthrough`
- `github.com/agentgateway/agentgateway/controller/api/v1alpha1/shared` — `LocalPolicyTargetReference`, `LocalPolicyTargetReferenceWithSectionName`

Upstream `controller/` is not a Go module on its own; `go get github.com/agentgateway/agentgateway/controller/api/v1alpha1/agentgateway@v1.2.1` pulls the parent module and the sub-packages resolve naturally. Go 1.25/1.26 compatibility verified.

Upstream field shape divergences from the old hand-rolled types: `McpTarget.Host` is `*ShortString` (pointer to string alias), `Path` is `*LongString`, `Protocol` is `*MCPProtocol`; `McpTargetSelector.Name` is `gwv1.SectionName`. `AgentgatewayPolicySpec.TargetRefs` is `[]shared.LocalPolicyTargetReferenceWithSectionName`. `BackendAuth.Passthrough` is `*BackendAuthPassthrough{}` (empty struct sentinel).

## DDD/hexagonal: ownerRef lives on the adapter

`Config` carries only domain identity (`Name`, `Namespace`). K8s-specific persistence fields (`UID`, `APIVersion`, `Kind` for `ownerReferences`) are bound on the Applier constructor — they don't leak into the domain. The yaml applier (PR 4) won't see them at all.

## Coverage

`go test -race -cover ./internal/reconciler/agentgateway/k8s/...` → **91.9%**. `golangci-lint run` clean. Whole-repo `go build ./...` clean.


QuentinBisson (commented):

DDD is clean — adapter imports domain, domain never imports `k8s/`. `Config` (k8s.Config, gateway name/ns) is correctly adapter-scoped.

**Blocking — correctness**
- `applyOwner` matches existing owner refs **by UID only**, then replaces. If the MCPServer is recreated with the same name and namespace, the new MCPServer has a new UID — `applyOwner` won't find the stale ref and will **append** the new one, leaving two `OwnerReferences` on the object. K8s GC then waits for both owners to disappear, so cascade delete silently breaks. The fake-client tests don't catch this because they never recreate. Match by `Name`+`Kind`+`APIVersion`, replace UID. Add a regression test for the recreate path.
- `createOrUpdate` re-fetches into the same `obj` on conflict, then re-runs the `mutate` closure which assigns `obj.Spec = …` wholesale — stomping any user-edited spec fields. Acceptable for `AgentgatewayBackend` (single-owner), but `HTTPRoute` is a Gateway-API object that users may legitimately modify (adding `filters`). Either narrow the spec mutation to fields you own or document "muster owns the entire HTTPRoute spec for these routes; user edits are reverted."

**Non-blocking — don't reinvent the wheel**
- `UpdateConflictRetries` defaulting to 3 immediate retries — use `retry.RetryOnConflict(retry.DefaultBackoff, fn)` from `k8s.io/client-go/util/retry`. Three immediate retries against a hot object dogpile etcd.
- Manual ownerRef field validation — `controllerutil.SetControllerReference(owner, controlled, scheme)` does it correctly and derives APIVersion/Kind from the scheme so callers don't pass them. Loses the per-Reconcile binding pattern but gains correctness.
- `applyOwner` doesn't set `BlockOwnerDeletion`; relies on caller's ownerRef carrying it. Fixture sets it; a future caller may not. Defensive construction or constructor-side derivation closes this.
- `mapProtocol` collapses any non-SSE protocol to `MCPProtocolStreamableHTTP`. If a future `HTTPProtocol` enum value lands (e.g., `WebSocket`), it silently mis-routes. Switch on the enum and return error for unknown values.
- `groupAgentgateway = "agentgateway.dev"` is hardcoded; use `agw.GroupVersion.Group` (or equivalent) — compile-time error on upstream rename instead of silent breakage.

**Tests**
- 91.9% is good. Gaps: owner-ref merge on a pre-existing object with a stale `OwnerReferences` entry (the blocker above); user-added `HTTPRoute.Spec.Rules[0].Filters` surviving re-apply (it doesn't); cascade test only checks the backend, not the HTTPRoute and Policy.

paurosello (approved):
LGTM. K8s emitter is clean — hexagonal split holds, ownerRef binds at constructor, reviewer feedback already addressed in-PR. Delete no-op is harmless until #707 wires it; #731 covers the suspend teardown.

PR #694, opened 2026-05-18
ci: consume go-build workspace binary; move build vars to pkg/project

QuentinBisson:
## Summary

Three coordinated changes that finish what architect-orb v8.1.0 simplification PR #690 started.

### 1. Dockerfile consumes the orb-built binary

`go-build` already produced a binary per arch but the Dockerfile re-compiled inside `docker buildx`. With `architectures: linux/amd64,linux/arm64` on `go-build`, both binaries land in the workspace; the Dockerfile is now a thin packaging step that `COPY`s the matching arch.

- Removes the duplicate compile and the QEMU-emulated arm64 cross-build on tag releases.
- `push-to-registries` auto-derives `--platform` from the workspace `.platforms` file (orb v8.1.0 feature).
- Branch CI keeps `platforms: linux/amd64` to skip the arm64 buildx leg for PR feedback speed.

### 2. Build identifiers move from `main` to `pkg/project`

Today the Dockerfile defines `ARG VERSION/COMMIT/DATE` and `-X main.version=…` ldflags, but the orb never passes those build args — so CI-built container images report `version=dev`. Goreleaser injects the same vars into `main`, and the orb's `go-test` step (which `go-build` runs first) populates `.ldflags` with `-X .../pkg/project.gitSHA=${CIRCLE_SHA1}` and `-X .../pkg/project.buildTimestamp=$(date --utc ...)` — but nothing in this repo reads those.

`pkg/project` now holds the canonical build vars. Both injection paths populate the same package:

| Path | Sets |
|---|---|
| goreleaser (release archives) | `version` (tag), `gitSHA` (short commit), `buildTimestamp` |
| architect-orb `go-build` (container images) | `gitSHA` (`CIRCLE_SHA1`), `buildTimestamp` (UTC) |

`muster version` prefers the tag → SHA → `dev`, and additionally prints `commit:` and `built:` lines so operators can pin a running CI image back to a commit.

### 3. Bump architect-orb pin to `8.2.1`

- **v8.2.1** ships a POSIX-shell fix for `image-login-to-registries`. v8.1.0 used bash-only syntax (`[[ ... ]]`, `${!var}`) that broke `sync-china-registry` (alpine/BusyBox executor) on its very first step — every `split-china-push: true` Aliyun mirror invocation was failing before any login attempt. This PR introduces exactly that topology, so the fix is load-bearing.
- **v8.2.0** flips `trimpath: true` to default-on for the `go-build` job, so the in-image binary picks up `-trimpath` without an explicit flag (byte-reproducible across hosts; stripped DWARF paths in `runtime.Caller` / panic traces / embedded build info). Goreleaser's own `-trimpath` flag stays as-is for release archives.


QuentinBisson (commented):

**Blocking — the stated goal isn't actually achieved**

PR body says: "Container images now correctly report the commit SHA instead of `dev`."

`architect-orb@8.1.0` `src/commands/go-build.yaml` only picks up ldflags from a `.ldflags` file in the workspace; otherwise it defaults to `-s -w` with no `-X` injection. No `.ldflags` file is added in this PR, and no CI step writes one. `Makefile.gen.go.mk` *does* set `pkg/project.gitSHA` / `pkg/project.buildTimestamp` (lines 22-24), but the orb invokes `go build` directly, not `make`.

Net effect: the orb-built `muster` binary still ships with `version=dev`, `gitSHA=dev`, `buildTimestamp=unknown`. The CHANGELOG entry is technically wrong as it stands.

Two fixes:

(a) add a `pre_test_target` Makefile target that writes `.ldflags` from `LDFLAGS` (with `BUILDTIMESTAMP`/`GITSHA1` wired from `CIRCLE_SHA1`), then expose it as a CircleCI step before `go-build`; or
(b) commit a `.ldflags` template and have a job step expand `${CIRCLE_SHA1}` into it.

Worth adding an actual CI smoke check: build the orb image, `muster version`, assert it doesn't print `dev`.

**Non-blocking**
- `gitSHA = dev` initial value uses the same `dev` constant as `version`. A 40-char SHA can never equal `"dev"`, so `Version()`'s `if gitSHA != dev` branch is safe — but document the contract in a comment on `gitSHA`.
- `buildTimestamp` is accessor-exposed but no caller is shown. If `muster version` is meant to print it, that wire-up is missing. If nothing reads it, the export is YAGNI.

**Tests**
- `main_test.go` correctly trimmed from 152 lines of tautological "setting a variable sets it" tests down to a focused `TestVersionWiring`. `pkg/project/project_test.go` covers the fallback table.

**Suggestion**
- Drop the narrative comment block (`# split-china-push: true delegates Aliyun ...`) once the workflow is settled — it's "why we did this", which is the kind of comment that ages badly.

paurosello (approved):

PR #695, opened 2026-05-18
feat(agentgateway): PR 4/16 — yaml emitter for filesystem mode

QuentinBisson:
Step **4/16** of the agentgateway adapter series. Plan: `/home/quentin/.claude/plans/declarative-forging-lynx.md`.

## What this lands

The YAML `Applier` — implementation of `agentgateway.Applier` (port from PR #689) for filesystem mode. One agw native YAML file per MCPServer, written atomically into a configured directory.

```go
applier, err := yaml.NewApplier(dir, yaml.WithListenerPort(8080))
if err != nil { ... }
if err := applier.Apply(ctx, config); err != nil { ... }
if err := applier.Delete(ctx, "name"); err != nil { ... }
```

Files:
- `internal/reconciler/agentgateway/yaml/doc.go`
- `internal/reconciler/agentgateway/yaml/applier.go` — `Applier`, `NewApplier`, `Apply`, `Delete`, `Close`, `Option`, `WithListenerPort`
- `internal/reconciler/agentgateway/yaml/local_types.go` — hand-rolled subset of the agw native config schema (`LocalConfig`, `LocalBind`, `LocalListener`, `LocalRoute`, `RouteMatch`, `PathMatch`, `LocalRouteBackend`, `LocalMcpBackend`, `LocalMcpTarget`, `McpTargetEndpoint`, `McpTargetStdio`, `FilterOrPolicy`, `BackendAuth`, `Passthrough`). Lives in the yaml package because it has exactly one consumer.
- `internal/reconciler/agentgateway/yaml/applier_test.go` — full coverage (round-trip, SSE/stdio variants, custom port, context cancellation, invalid names, malformed configs, stale-content overwrite, leftover-tmp cleanup, delete lifecycle, concurrent distinct names, golden snapshots)
- `internal/reconciler/agentgateway/yaml/testdata/grizzly.yaml` + `grizzly-stdio.yaml` — golden files (regenerate with `UPDATE_GOLDEN=1 go test ./...`)

## Atomic write contract

`Apply` is **byte-identical** on re-apply: if the on-disk file already matches the new payload, the write is skipped. When content does differ, write goes to `<name>.yaml.tmp` → `fsync` → `rename` over `<name>.yaml`. Any preexisting `.yaml.tmp` from a crashed prior write is cleaned by the rename. Concurrent applies for distinct names parallelize via a per-name mutex map; same-name applies serialize. `Delete` drops its entry from the lock map under the global mutex so distinct-name churn does not leak `*sync.Mutex` entries.

Name validation uses `k8s.io/apimachinery/pkg/util/validation.IsDNS1123Subdomain`, so the applier cannot be coerced into writing outside its configured `os.Root`.

## Domain helpers (also consumed by k8s adapter)

Two exported helpers on the agentgateway domain config:

- `Authn.RequiresPolicy() bool` — true when the gateway needs a policy resource (`Type != None` or `ForwardToken`).
- `HTTPTarget.Validate() error` — port range check, called from `NewConfig` so downstream adapters can drop their duplicates.

`internal/reconciler/agentgateway/k8s/reconcile.go` is the one cross-cutting touch: `reconcileBackend` now calls `target.Validate()` at the top so the cluster-mode adapter shares the same precondition as the yaml adapter.

## Why hand-rolled types, not go-jsonschema codegen

Codegen via `github.com/atombender/go-jsonschema` was attempted and failed with:

```
cannot add struct field: could not generate type with scope 'AzureContentSafetyPolicies':
  cannot add struct field: could not generate type with scope 'AzureContentSafetyPoliciesBackendAuth':
    could not merge anyOf types: types list is empty
```

The agw schema uses **untagged JSON `oneOf` for Rust serde-style externally-tagged enums**:

```jsonc
"BackendAuth": {
  "oneOf": [
    { "type": "string", "enum": ["copilot"] },
    { "type": "object", "properties": { "passthrough": {…} }, "required": ["passthrough"] },
    { "type": "object", "properties": { "key": {…} }, "required": ["key"] },
    { "type": "object", "properties": { "gcp": {…} }, "required": ["gcp"] },
    { "type": "object", "properties": { "aws": {…} }, "required": ["aws"] },
    { "type": "object", "properties": { "azure": {…} }, "required": ["azure"] }
  ]
}
```

Mixing a string variant with object variants in a single `oneOf` is a Rust enum representation (`#[serde(untagged)]` / externally-tagged). Go's type system has no native sum type — faithful codegen would need an interface plus per-variant types and a hand-written `UnmarshalJSON`. `go-jsonschema` v0.23.1 instead tries to merge the variants into a single struct, finds incompatible top-level types, and aborts.

Hand-rolling is in fact the right answer regardless: muster emits **only** `mcp.targets[].{mcp|sse|stdio}` and `policies.backendAuth.passthrough`. Even working codegen would pull in thousands of lines of AI providers, AWS/Azure/GCP auth, transformations, etc. that muster never serializes. The types live in the yaml package (~115 LOC); the package doc records the rationale and pinned schema URL.

## Incidental: gosec G101 noise moved inline

`.golangci.yml`'s four per-file `gosec`/`G101` exclude blocks (for `internal/events/types.go`, `pkg/oauth/wire.go`, `internal/testing/mock/*.go`) are removed. Each false positive is now annotated at the call site with a `//nolint:gosec // G101: ...` comment naming the RFC / fixture context. Same outcome, but the suppression and its justification sit next to the flagged literal instead of in a central config file. Touched files: `.golangci.yml`, `internal/events/types.go`, `internal/testing/mock/oauth_server.go`, `pkg/oauth/wire.go`. Unrelated to the yaml emitter; bundled here because the rebase put them on the same branch.

## Coverage

`go test -race -cover ./internal/reconciler/agentgateway/yaml/...` → **83.3%** (uncovered: writeAtomic os.Sync/Rename failure branches that require real-FS fault injection). `go build ./...` and `golangci-lint run` clean.


QuentinBisson (commented):

DDD is clean — imports `agentgateway` (domain) and stdlib only. `LocalConfig` and friends are correctly local serialization shapes.

**Blocking — stale doc reference**
- `doc.go` references `internal/agentgateway/nativeconfig` for the Go shape — that path doesn't exist; the types live in this package as `local_types.go`. Fix before merge; exactly the kind of stale path that survives forever.

**Non-blocking**
- `lockFor` grows `a.locks` unboundedly — every distinct MCPServer name ever applied leaks a `*sync.Mutex` until `Close()`. For a long-lived controller across hundreds of churned MCPServers, real. Either `delete(a.locks, name)` in `Delete` (under the global mu) or `sync.Map` with documented bound.
- `nameSafe` regex is essentially RFC 1123 DNS subdomain. Use `k8s.io/apimachinery/pkg/util/validation.IsDNS1123Subdomain` or document the deliberate divergence — hand-rolled K8s name regex is a known foot-gun (Unicode, trailing dots).
- `gopkg.in/yaml.v3` is maintenance-mode; `sigs.k8s.io/yaml` is the K8s-ecosystem standard and is already on go.mod. Trade-off real (sigs goes through JSON tags, not yaml tags) — flag consciously rather than default.
- `Apply` reads the existing file via `ReadFile` and skips write on `bytes.Equal`. Fine at ~1KB; document the budget assumption.
- `policyFor` only sets `BackendAuth.Passthrough` for `ForwardToken`. Silently drops `RequiredAudiences`, `TokenExchange`, `AuthorizationServer`. If filesystem-mode-supported later, leave a TODO; if intentionally cluster-only, document on `agentgateway.Authn`.

**Tests**
- 83.3% with golden files is appropriate. Missing: concurrent same-name applies; `Close` followed by `Apply` returning `os.ErrClosed`; deterministic output with non-empty `RequiredAudiences`; a 2-env-var stdio golden to confirm `yaml.v3` map-key sort order (it sorts, but lock it).

**Series cross-cuts** (with #692)
- `policyFor` (yaml) and `policySpec` (k8s) both gate on `Authn.Type == None && !ForwardToken`. Hoist to `(Authn).RequiresPolicy() bool` on the domain — single source of truth, both adapters call.
- HTTP port range validation (`portInt32` in k8s/helpers.go; inline in yaml/applier.go's `httpEndpoint`). Hoist to `agentgateway.HTTPTarget.Validate()` called from `NewConfig`.
- Name-equals-backend-equals-route-equals-policy invariant is enforced in yaml (`nameFromConfig`) but NOT in the K8s applier. Lift to `Config.Validate()`.

PR #699, opened 2026-05-18
feat(musterstdio): stdio→MCP-over-StreamableHTTP shim binary (PR 5/10)

QuentinBisson:
**Step 5 of the 10-PR translator series** ([plan](../CLAUDE.md)).

Standalone Go binary that wraps an stdio MCP child as MCP-over-StreamableHTTP. It is the shim image referenced by the translator stack for `spec.type: stdio` MCPServer specs — filesystem mode (PR 7+) will run it as a local subprocess, cluster mode (PR 3) as a `Deployment`+`Service`.

## What's in

| Path | Purpose |
|---|---|
| `cmd/musterstdio/main.go` | `Run(ctx, args, stderr, logger)` + `main()` wired to `signal.NotifyContext` for SIGTERM/SIGINT |
| `cmd/musterstdio/server.go` | HTTP listener: `POST /mcp`, `GET /healthz`, optional separate health port, drain-on-shutdown |
| `cmd/musterstdio/bridge.go` | `os/exec` child + JSON-RPC frame demuxer keyed on the `id` field; SIGTERM → wait → SIGKILL fallback |
| `cmd/musterstdio/*_test.go` | Unit + scenario + end-to-end + goroutine-leak tests (≥80% coverage; race-clean) |
| `Dockerfile.musterstdio` | Distroless multi-arch (`linux/amd64`, `linux/arm64`) image |
| `.goreleaser.yaml` / `Makefile.custom.mk` | Release pipeline + `make musterstdio-image[-multiarch]` targets |
| `internal/reconciler/translator/shimimage.go` | Single `ShimImage` constant pinned to the muster version |

## Behaviour

- Flags: `--child-command`, `--child-arg` (repeatable), `--child-env KEY=VAL` (repeatable), `--listen-port` (default 8080), `--health-port` (optional, separate listener), `--shutdown-timeout`, `--startup-timeout`, `--max-request-bytes`, `--version`
- `POST /mcp` reads the body as a JSON-RPC frame, writes it as a `\n`-terminated line to the child's stdin, and waits for the response frame carrying the same `id`. Notifications (no `id`) return `202 Accepted`.
- `GET /healthz` reports `200` while the child is running and stdout has not errored, `503` otherwise.
- Concurrent POSTs are multiplexed by JSON-RPC `id`; out-of-order responses from the child are routed back to the correct caller.
- SIGTERM stops accepting new HTTP requests, drains in-flight Sends, closes the child's stdin, sends SIGTERM, waits up to `--shutdown-timeout`, then SIGKILLs.
- Child crash → `Run` returns a wrapped error → exit non-zero with a diagnostic.

## Tests

- `bridge_test.go` — request/response, notification, concurrent out-of-order multiplexing, context-cancel cleanup, child exit, SIGKILL fallback, invalid frame.
- `server_test.go` — happy path, notification 202, bridge-error 502, unhealthy 503, empty/invalid body 400, method-not-allowed 405, in-flight drain on shutdown, separate health listener.
- `scenarios_test.go` — Given/When/Then BDD-style end-to-end coverage against the in-process fake stdio child (CRD-driven YAML scenarios for the shim land in PR 6/9 once the reconciler wires it through `MCPServer`).
- `main_test.go` — flag validation, `--version`, in-process `Run` lifecycle, real-subprocess SIGTERM clean-exit, real-subprocess child crash, `goleak.VerifyNone` after shutdown.

Coverage: **81.9%** on the new package. `make test -race` clean. `make lint` introduces 0 new findings (verified against the PR 4 baseline).

## Lint

A tightly scoped `.golangci.yml` exemption covers `gosec G204` in the three files that actually exec subprocesses (`bridge.go` — the whole point of the binary; `main_test.go` and `testhelpers_test.go` — test infra). Everything else under `cmd/musterstdio/` is still scanned.

## Sample HTTP trace

```text
$ musterstdio --child-command /tmp/demo-child.sh --listen-port 0
time=2026-05-18T18:25:49 level=INFO msg="musterstdio ready" version=dev listen_addr=[::]:40777 child=/tmp/demo-child.sh

$ curl -i http://[::]:40777/healthz
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 15

{"status":"ok"}

$ curl -i -H 'Content-Type: application/json' \
       -d '{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}' \
       http://[::]:40777/mcp
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 61

{"jsonrpc":"2.0","id":1,"result":{"tools":[{"name":"echo"}]}}

$ curl -i -H 'Content-Type: application/json' \
       -d '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
       http://[::]:40777/mcp
HTTP/1.1 202 Accepted
Content-Length: 0

$ kill -TERM <pid>   # clean exit, no leaked goroutines
```

## LOC

| | Lines |
|---|--:|
| Added (production) | ~790 (`bridge.go` 331 + `server.go` 245 + `main.go` 201 + `doc.go` 6 + `shimimage.go` 14) |
| Added (tests) | ~1,210 |
| Build / release wiring | `Dockerfile.musterstdio` (33), `.goreleaser.yaml` (+~30), `Makefile.custom.mk` (+~15) |

All production + test files are ≤400 LOC.

## Out of scope (deferred)

- JSON-RPC batch demuxing — single frame in / single frame out is the MCP stdio shape today; batch support lands when an actual client needs it.
- Server-to-client notification streaming — current bridge drops unmatched frames; the StreamableHTTP spec supports it via SSE upgrade, which a future PR can layer over `bridge.deliver`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ logs
 stress-test-results/
 .klaus/
 muster-v-*
+musterstdio
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,3 +14,18 @@ linters:
       - path: _test\.go
         linters: [gosec]
         text: G101
+      # musterstdio is by design a generic stdio→HTTP shim: its --child-command
+      # / --child-arg flags carry user-controlled values straight to os/exec.
+      # G204's heuristic flags every exec.Command call in these files. Only the
+      # specific files that actually need to launch the configured child (or,
+      # in the test helpers, the built shim binary / re-exec'd fake child) are
+      # exempt — the rest of cmd/musterstdio is still scanned.
+      - path: cmd/musterstdio/bridge\.go
+        linters: [gosec]
+        text: G204
+      - path: cmd/musterstdio/main_test\.go
+        linters: [gosec]
+        text: G204
+      - path: cmd/musterstdio/testhelpers_test\.go
+        linters: [gosec]
+        text: G204
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -50,9 +50,28 @@ builds:
       - >-
         -s -w -X main.version={{.Version}} -X main.commit={{.ShortCommit}}
         -X main.date={{.Date}}
+  - id: musterstdio
+    main: ./cmd/musterstdio
+    binary: musterstdio
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
+    env:
+      - CGO_ENABLED=0
+    mod_timestamp: '{{ .CommitTimestamp }}'
+    flags:
+      - -trimpath
+    ldflags:
+      - >-
+        -s -w -X main.version={{.Version}} -X main.commit={{.ShortCommit}}
+        -X main.date={{.Date}}
 
 archives:
   - id: default
+    builds:
+      - muster
     # Default formats for all platforms (can be overridden per OS)
     formats: [tar.gz]
     name_template: >-
@@ -74,6 +93,18 @@ archives:
     files:
       - LICENSE
       - README.md
+  - id: musterstdio
+    builds:
+      - musterstdio
+    formats: [tar.gz]
+    name_template: >-
+      musterstdio_ {{- title .Os }}_ {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "386" }}i386
+      {{- else }}{{ .Arch }}{{ end }}
+      {{- if .Arm }}v{{ .Arm }}{{ end }}
+    wrap_in_directory: true
+    files:
+      - LICENSE
 
 checksum:
   # Algorithm to use. Valid options are sha256, sha512, sha1, crc32, md5, etc.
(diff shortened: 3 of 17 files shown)


QuentinBisson:
Superseded — agentgateway v1.2.1's `LocalMcpTarget` natively supports a `stdio` variant (`cmd`, `args`, `env`, `clear_env`), confirmed against the [v1.2.1 schema](https://raw.githubusercontent.com/agentgateway/agentgateway/refs/tags/v1.2.1/schema/config.json) ($defs.LocalMcpTarget). agentgateway itself spawns the child and demuxes JSON-RPC, so a separate musterstdio shim is unnecessary in filesystem mode. Cluster-mode stdio support is deferred — the initial cluster release supports streamable-http and sse MCPServers only, and stdio in cluster mode will be revisited when isolation requirements are clear.

Follow-up edits:
- PR 1 (#688): drop `ShimRequest`/`Endpoint`/`Model.Shims`; `Backend` becomes a tagged union covering stdio directly.
- PR 3 (#692): drop stdio shim Deployment+Service emission; cluster-mode stdio sets a `NotSupportedInCluster` status condition.
- PR 4 (#695): emit `mcp.targets[].stdio` for stdio backends.
- Plan: rewritten locked-decisions row + PR 5/6/7 sections.

Issue #637, opened 2026-05-07
Pre-commit goconst lint debt blocks PRs after golangci-lint 2.12.2 bump


teemow:
## Problem

After Align files PR #629 lands (bumps `golangci-lint` from 2.11.4 to 2.12.2), the `pre-commit` job reports **947 new `goconst` violations** across `cmd/`, `internal/`, `pkg/`, and `main_test.go`. Sample:

- `cmd/agent_test.go:34: string \`text\` has 18 occurrences`
- `cmd/call_test.go:26: string \`name\` has 15 occurrences`
- `pkg/oauth/client_test.go:53: string \`https://issuer.example.com\` has 9 occurrences`
- `pkg/oauth/pkce.go:32: string \`S256\` has 5 occurrences`

The bulk of the violations are repeated test fixtures (`Bearer`, `S256`, `text`, URLs) that goconst v1.8.2 didn't flag but v1.10.0 (shipped in golangci-lint 2.12.0) does -- see giantswarm/devctl#1802 for the upstream regression.

Until the `pre-commit` job is green, every Renovate / Align files PR in muster is blocked.

## Proposed solution

Two paths, pick one (or both):

1. **Add a per-repo `.golangci.yml` that silences the noise**:

   ```yaml
   linters-settings:
     goconst:
       min-len: 5
       min-occurrences: 5
       ignore-tests: true
   ```

   This lets goconst still flag genuine production-code duplication while ignoring the test-fixture repetition that drives almost all 947 violations.

2. **Fix the genuine production duplications** that survive the tuning above. Most are in `cmd/` (auth subcommand names, JSON keys) and `pkg/oauth/` (`S256` / `Bearer` / OAuth header keys). Extract them to package-level constants -- following the pattern from giantswarm/lab#77.

The Align files PR (#629) cannot land until `pre-commit` passes. Either the per-repo `.golangci.yml` workaround above, or upstream resolution of giantswarm/devctl#1802 (which would tune the default), unblocks #629.

## Acceptance criteria

- `pre-commit run --all-files` exits 0 on `main`.
- Align files PR #629 (or its successor after the workaround/upstream fix lands) merges cleanly.
- The chosen `goconst` configuration is documented in the repo (`.golangci.yml` and/or a comment) so future contributors understand why this rule is tuned.

## References

- giantswarm/muster#629 -- the blocked Align files PR.
[...]

PR #629, opened 2026-05-06
Align files

architectbot:
This PR is generated by https://github.com/giantswarm/github, it will be merged automatically once all requirements are met

PR #700, opened 2026-05-18
chore(lint): scope goconst/gosec via .golangci.yml and fix gosec findings

QuentinBisson:
## Summary

Unblocks #629 (Align files) and resolves the lint debt tracked in #637.

`.golangci.yml` (new):
- `goconst`: `min-len: 5`, `min-occurrences: 5`, `ignore-tests: true`, plus an `ignore-functions` list covering muster's logging helpers and `fmt.Errorf` so log/error format strings stop firing as duplicates.
- `gosec`: excluded from `_test.go` files (test fixtures contain repeated dummy tokens by design); `G101` excluded narrowly on `internal/events/types.go` where event-reason identifiers contain the word ``Token``.

Source-level fixes (no nolint substitutes):
- `cmd/`: lifted resource-type / MCP primitive identifiers to constants in `cmd/mcp_types.go`; the existing `//nolint:goconst` directives on those strings are now redundant.
- `internal/aggregator/`: added `constants.go` with shared `defaultValkeyKeyPrefix`/`httpReadHeaderTimeout`; switched the default namespace to `metav1.NamespaceDefault`; replaced the running/connected string comparisons with `api.IsActiveState` and the corresponding `api.HealthHealthy`/`api.HealthUnhealthy` constants in the event handler; set `ReadHeaderTimeout` on every HTTP server construction (fixes G112).
- `internal/context/storage.go`: `filepath.Clean` on the user-supplied config path; tightened directory/file permissions to `0o700`/`0o600` (G301/G306/G304).
- `internal/oauth/manager.go`: `filepath.Clean` on the configurable CA-file path (G304).
- `internal/orchestrator/orchestrator.go`: dropped a stale `//nolint:gosec` on a `context.WithCancel` call that gosec no longer flags.
- `internal/aggregator/server.go`: dropped a stale `//nolint:gosec` on the SSO init goroutine spawn.
- `internal/events/types.go`: dropped per-line `//nolint:gosec` on the `MCPServerToken*` event reasons; now covered by the file-level exclusion above.
- Stripped 152 redundant `//nolint:goconst,gosec` directives from `_test.go` files (now covered by the config).

Result: `golangci-lint run -E=gosec -E=goconst -E=govet` reports a single residual category (workflow/cli/mcpserver JSON-schema vocabulary), which is addressed in the follow-up PRs in this stack (`api.ArgType` enum, shared resource-type constants, JSON-key constants).

## References
- #629
- #637


QuentinBisson (commented):

`0o644 → 0o600` on user config and `0o755 → 0o700` on dirs are correct G306 fixes. No new `#nosec` annotations in source — only ones removed — which is the right direction.

**Non-blocking — scope creep beyond "lint fix"**
- `event_handler.go` swaps inline string equality (`event.NewState == "running" || "connected"`) for `api.IsActiveState(...)`, `api.HealthHealthy`, `api.StateStarting`, etc. That's a *behavior refactor*, not a gosec fix. Behavior-equivalent only if `IsActiveState` covers exactly `{running, connected}` and the typed-string conversions are loss-less. Either surface in the PR body or split into its own PR — the title says "fix gosec findings".

**Non-blocking — wrong-layer constant**
- `internal/aggregator/constants.go` introduces `defaultValkeyKeyPrefix`, but the value originates in `internal/config`. PR #706 then moves it. The factoring here is at the wrong layer and gets reverted next PR — net churn.

**Non-blocking — `.golangci.yml`**
- The file-level scoping (`internal/aggregator/server.go`, `internal/admin/handlers.go`) hides any *new* G118/G710 in those files until #706 narrows them. Add an issue link in the YAML comment so the exclusion has a sunset.
- The 152 redundant `//nolint` cleanups in `_test.go` rely on the new path-based test exclusion — confirm in CI that goconst's `ignore-tests` setting alone wouldn't have sufficed (one source of truth instead of two).

paurosello (approved):
Conflicts resolved by merging main into the branch — verified `go build ./...`, full `go test ./internal/aggregator/...` (4.4s, all green), and `make lint` (0 issues). The `RegisterPendingAuth` refactor on main was accommodated by adopting the new `PendingAuthRegistration` struct signature; the `//nolint:gosec` annotations from main's tests are now redundant given the `_test.go` path exclusion this PR adds to `.golangci.yml`.

PR #701, opened 2026-05-18
refactor(api): introduce ArgType enum for ArgMetadata.Type

QuentinBisson:
## Summary

Stacked on #700.

`api.ArgMetadata.Type` was a stringly-typed field whose valid values (\`string\`, \`number\`, \`integer\`, \`boolean\`, \`object\`, \`array\`) were documented only in a comment. This PR introduces a typed `api.ArgType` enum with one constant per valid value, then converts every adapter that produces `ArgMetadata` (workflow, mcpserver, orchestrator, events, app, metatools) and the `validateType` switch in `internal/workflow/validation.go` to use it.

Wire format is unchanged: `ArgType` marshals identically to its underlying string.

## What this fixes
- Compiler-enforced vocabulary at every call site instead of a docstring contract.
- Removes one whole class of goconst hits (`type: "string"`, `type: "object"`, etc., 70+ occurrences) without requiring `ignore-string-values` for them.


QuentinBisson:

PR #702, opened 2026-05-18
refactor(cli,api): share resource-type and JSON-key constants

QuentinBisson:
## Summary

Stacked on #701.

Lifts the resource-type identifiers (\`service\`, \`mcpserver\`, \`workflow\`, \`workflow-execution\`, ...) from \`cmd/mcp_types.go\` to \`internal/api/resource_types.go\` so \`internal/cli/table_formatter.go\` and friends reference the same source of truth instead of duplicating the literal strings.

Adds two more constants packages alongside:
- \`internal/api/json_keys.go\` — JSON-Schema spec keys (\`description\`, \`properties\`, \`items\`, \`required\`, \`default\`, \`enum\`, \`additionalProperties\`) and muster API response field names (\`name\`, \`status\`, \`state\`, \`health\`, \`command\`, \`mimeType\`, \`inputSchema\`, ...).
- \`internal/cli/columns.go\` — kubectl-style table header labels (\`headerName\`, \`headerDescription\`, ...).

Sweeps \`internal/cli/{table_formatter,table_builder,mcp_output}.go\` to reference the named constants instead of repeated string literals. Applied consistently to every occurrence in those files, not only the ones goconst flagged, so the \`description\`/\`name\` field-name vocabulary has a single canonical symbol rather than a mix of literals and constants.

\`cmd/mcp_types.go\` keeps short package-local aliases (\`resourceTypeService = api.ResourceTypeService\`, etc.) so the existing \`cmd/*.go\` switches and slice literals stay terse.

### DIFF
--- a/cmd/mcp_types.go
+++ b/cmd/mcp_types.go
@@ -1,21 +1,26 @@
 package cmd
 
-const (
-	resourceTypeService            = "service"
-	resourceTypeServices           = "services"
-	resourceTypeMCPServer          = "mcpserver"
-	resourceTypeMCPServers         = "mcpservers"
-	resourceTypeWorkflow           = "workflow"
-	resourceTypeWorkflows          = "workflows"
-	resourceTypeWorkflowExecution  = "workflow-execution"
-	resourceTypeWorkflowExecutions = "workflow-executions"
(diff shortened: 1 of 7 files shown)


QuentinBisson:

PR #703, opened 2026-05-18
refactor: extract JSON Schema, OAuth wire, and workflow op constants

QuentinBisson:
## Summary

Stacked on #702.

Removes the remaining duplicated string literals flagged by goconst without resorting to \`ignore-string-values\`. After this PR the \`.golangci.yml\` no longer needs that escape hatch.

- \`pkg/oauth/wire.go\` — RFC 6749/6750/7636/8693 vocabulary (\`HeaderAuthorization\`, \`SchemeBearer\`, \`FormField*\`, \`GrantType*\`, \`CodeChallengeMethod*\`, \`JSONField*\`, \`Err*\`). Adopted by \`internal/aggregator/{server,connection_helper}.go\` and the mock OAuth server, replacing scattered \`Authorization\`/\`Bearer\`/\`invalid_grant\`/\`error_description\` literals.
- \`internal/workflow/operations.go\` — \`opExecute\`/\`opValidate\` for workflow lifecycle event \`Operation\` strings; applied across \`api_adapter.go\`.
- \`internal/agent/commands/constants.go\` — \`boolStringTrue\`/\`boolStringFalse\` for stringly-typed boolean parsing in filter/notifications commands.
- \`internal/workflow/{api_adapter,executor,execution_tracker}.go\`, \`internal/mcpserver/api_adapter.go\`, \`internal/metatools/{handlers,formatters}.go\`, \`internal/testing/{test_tools,muster_manager}.go\` — JSON-key literals swapped for \`api.SchemaKey*\`/\`api.Field*\` constants. Applied consistently rather than only on goconst hot spots.
- \`internal/testing/test_tools.go\` — extracted the \`default\` user identifier to a \`defaultTestUser\` constant.

Remaining gosec/govet noise:
- \`internal/admin/handlers.go\`: \`url.PathEscape\` the session id before redirecting (G710 open-redirect taint flow).
- \`internal/aggregator/server.go\`: pass \`context.Background()\` to the SSO bootstrap goroutine since it must outlive the request handler; narrow \`//nolint:gosec\` with rationale (G118).
- \`internal/testing/muster_manager.go\`: \`reflect.Ptr\` → \`reflect.Pointer\`.
- \`.golangci.yml\`: drop \`ignore-string-values\`; add path-based exclusions for \`pkg/oauth/wire.go\` (G101 false positive on the token-exchange URN) and \`internal/testing/mock/*\` (G710 on the BDD mock OAuth server, which must honour \`redirect_uri\`).

[...]

PR #704, opened 2026-05-18
refactor(config): share DefaultValkeyKeyPrefix across aggregator and oauth

QuentinBisson:
## Summary

Stacked on #703.

The \`muster:\` Valkey key prefix was duplicated as a local default in four Valkey-backed stores:
- \`internal/aggregator/capability_store_valkey.go\`
- \`internal/aggregator/session_auth_store_valkey.go\`
- \`internal/oauth/token_store_valkey.go\`
- \`internal/oauth/state_store_valkey.go\`

Plus the aggregator HTTP-server constants file held a third copy as \`defaultValkeyKeyPrefix\` (introduced in #700).

Lifts the constant to \`internal/config.DefaultValkeyKeyPrefix\`, adjacent to the existing \`ValkeyConfig.KeyPrefix\` field that already owns the operator override. Every store now derives the same default from one symbol, so changing the namespace requires touching exactly one line.

\`internal/aggregator/constants.go\` keeps only \`httpReadHeaderTimeout\`.

### DIFF
--- a/internal/aggregator/capability_store_valkey.go
+++ b/internal/aggregator/capability_store_valkey.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/valkey-io/valkey-go"
 
+	"github.com/giantswarm/muster/internal/config"
 	"github.com/giantswarm/muster/pkg/logging"
 )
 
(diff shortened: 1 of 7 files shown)


QuentinBisson:

Issue #698, opened 2026-05-18
spec.family — fail-fast on instanceArg conflict (design)


QuentinBisson:
Design issue tracking the option to enforce \`family.instanceArg\` uniqueness at admission / reconciler time, rather than continue with the graceful-fallback behaviour in #670.

Split out of #697 because muster's existing reconcilers only handle Kubernetes optimistic-locking conflicts — there is no precedent in this codebase for cross-CR domain-conflict detection or \`Ready=False, reason=…\` patterns. Introducing one is a new pattern, not a follow-up bug fix.

## Current behaviour (#670)

When two \`MCPServer\` CRs declare the same \`spec.family.name\` with different \`spec.family.instanceArg\`, the aggregator detects divergence at the family level (via \`familyFallbackStatusLocked\` in \`internal/aggregator/registry.go\`) and falls back the entire family to per-server prefixing. A single warning is logged at the moment the divergent server registers.

Operator UX:
- Both CRs remain \`Ready=True\`.
- No \`x_<family>_*\` tools surface.
- The warning is only visible in aggregator logs.

## Proposed alternative

Refuse to register the second CR. Surface \`Status.Conditions[type=Ready, status=False, reason=FamilyInstanceArgConflict]\` with a message identifying the offending sibling. Tiebreaker: older \`creationTimestamp\` wins; later CR enters \`Ready=False\` until the conflict is resolved.

## What this depends on

1. **Design decision** — is the new pattern worth introducing in muster? Trade-offs:
   - Operator UX: \`Ready=False\` + reason is clearly better than a log message.
   - Implementation cost: new pattern, more code, more state to manage.
   - Multi-tenant safety: with fail-fast a misconfigured CR cannot silently disable a family for everyone.
2. **\`api.MCPServerInfo\` extension** — \`internal/mcpserver/api_adapter.go::convertCRDToInfo\` does not currently surface \`Status.Conditions[]\`. Adding the field is mechanical (~20 lines) but only useful if there is a condition to observe.
3. **Reconciler enforcement** — add cross-CR \`List\` + tiebreaker logic in \`internal/reconciler/mcpserver_reconciler.go\`. Defence-in-depth at \`aggregator.Register\` returning \`ErrFamilyInstanceArgConflict\`.
4. **BDD scenarios** — two scenarios from the original #697 list become writeable once \`api.MCPServerInfo.Conditions\` exists:
[...]

PR #705, opened 2026-05-18
fix(family): deep-copy nested schemas, stable tools/list ordering, instanceArg collision fallback

### LINKED ISSUE #697: spec.family — deferred bug surfaces and behavioural reproducers
Tracks behaviour gaps in `MCPServer.spec.family` that are either real bugs without an actionable reproducer in the current BDD framework, or design questions whose resolution unblocks new scenarios. Companion to #670 and to the two intentional bug reproducers already added in that PR (`mcpserver-family-cross-session-corruption`, `mcpserver-family-disjoint-tools-divergence`).

### 1. `injectInstanceEnum` deep-copy is shallow
`internal/aggregator/registry.go` lines ~547–570 clone only top-level `InputSchema.Properties` and `Required`. Nested object `properties`/`required` and array `items` schemas are still shared with the cached per-server tool. Mutations through repeated `GetAllTools` calls can leak across calls when family tools expose nested schemas.

**Surface:** add a unit test in `internal/aggregator/registry_test.go` that builds a family tool with a nested object property, calls `GetAllTools` twice, and verifies the second call returns a pristine nested schema. No BDD scenario — the failure mode isn't observable through tool-call results alone.

### 2. `assembleExposedTools` tool ordering is non-deterministic
`familyBuckets` is iterated as a Go map, so `tools/list` ordering varies per call. Could cause flaky `tools/list_changed` diffs downstream.

**Surface:** unit test that calls `assembleExposedTools` N times and asserts stable order (lexical by exposed name is the simplest deterministic rule). No BDD scenario — the framework's `expected.contains` is substring-only; no ordering assertions today.

### 3. Concurrent `GetAllToolsForSession` race
`r.nameMu` protects the write-back to `familyMappings`, but the contribution-collection phase before the lock is unprotected. Two sessions resetting + rewriting the family routing index in rapid succession can leave a third session reading a half-built index.

**Surface:** unit test with `sync.WaitGroup` and `-race`, two goroutines hammering `GetAllToolsForSession` with disjoint authenticated server sets, asserting no `panic` and consistent `CallToolInternal` resolution per session. The BDD framework runs steps sequentially — no parallel-step support in `internal/testing/types.go::TestStep`. A behavioural reproducer would need framework changes.

### 4. `family.instanceArg` name collision with a tool-defined input property
Today `injectInstanceEnum` silently overwrites the colliding property. The tool's original schema (description, type, freedom) is lost, and the aggregator consumes the value as routing data so the backend never sees it.

Three viable fix paths, each implying a different test assertion:
- **(a) Reject family registration on collision; fall back to per-server prefixing for that server.** Mirrors how description-divergence already falls back. Test would assert `x_<server>_<tool>` is exposed with the original schema intact.
- **(b) Reject at admission / reconciler condition.** `Status.Conditions.Ready=False, reason=FamilyInstanceArgConflict`. Today blocked by item 6 below — `core_mcpserver_get` returns `api.MCPServerInfo` which does not expose `Status.Conditions[]`.
- **(c) Rename the injected routing arg out of the way.** Both the original `cluster` (tool meaning) and a renamed routing arg co-exist on the family-prefixed tool.

A scenario draft anchoring on (a) was prepared but deferred pending this decision.

### 5. `family.instanceArg` mismatch across CRs in the same `family.name` — fail-fast vs graceful fallback
Today the aggregator silently falls back when two CRs declare the same `family.name` with different `family.instanceArg` (`mcpserver-family-instance-arg-divergence` passes against this graceful behaviour). The disjoint-tools reproducer added in #670 (`mcpserver-family-disjoint-tools-divergence`) shows the fallback is also incomplete (per-(family, tool) granularity misses family-level divergence on disjoint tool sets).

Two design directions:
- **Two-pass aggregator refactor.** First pass validates family invariants across all contributors; second pass groups tools. Family-level fallback runs uniformly regardless of tool overlap. Fixes the disjoint-tools bug at the aggregator layer.
- **Fail-fast at the reconciler.** Sets `Ready=False, reason=FamilyInstanceArgConflict` on the conflicting CR (tiebreaker: older `creationTimestamp` wins). The conflicting CR never reaches the aggregator. Eliminates the bug class structurally; depends on item 6 to write the scenario.

Recommendation in the planning notes was to do both — reconciler-level fail-fast as primary, two-pass refactor as defence-in-depth — but this needs maintainer sign-off.

### 6. `api.MCPServerInfo` does not surface `Status.Conditions[]`
`internal/mcpserver/api_adapter.go::convertCRDToInfo` exposes `State`, `LastError`, `ConsecutiveFailures` but not the full conditions list. Blocks any BDD scenario that needs to observe reconciler-set conditions (items 4(b), 5 reconciler path, future status-driven scenarios).

**Surface:** add `Conditions []MCPServerCondition` to `api.MCPServerInfo` and have `core_mcpserver_get` JSON-marshal them so scenarios can json-path-assert `.conditions[?(@.type==\"Ready\")].reason`.

## Deferred BDD scenarios (unblocked once the above decisions land)

- **`mcpserver-family-instance-arg-name-collision`** — depends on #4 fix-path decision.
- **`mcpserver-family-instance-arg-conflict-reconciler`** — depends on #5 reconciler fail-fast + #6 condition exposure.
- **`mcpserver-family-instance-arg-conflict-resolution-on-delete`** — same dependencies as above; asserts the loser becomes Ready after the winner is deleted.

## Scenarios already added in this work

- `mcpserver-family-three-member-partial-deregister` — regression guard, broader than the existing two-member deregister coverage. Passes today.
- `mcpserver-family-partial-description-divergence-strict` — pins the negative assertion that the diverging tool's family-prefixed name is absent. Passes today.

## Out of scope
A separate `MCPServerFamily` CRD is mentioned as a possible future shape if the family surface grows beyond `name + instanceArg` (per #670 description's follow-ups section). Not blocking any of the above.

QuentinBisson:
Stacked on top of #670. Closes the actionable items from #697 — leaves the design questions (fail-fast at admission, MCPServerInfo.Conditions exposure) for #698.

### 1. \`injectInstanceEnum\` deep-copy was shallow

\`Properties\` and \`Required\` were cloned at the top level, but the nested map values inside \`Properties\` (object sub-schemas, array \`items\`, nested \`required\`) were shared by reference with the backend's cached tool. A caller walking the exposed schema and mutating a nested map leaked the mutation back into the per-server cache, breaking later \`GetAllTools\` calls.

\`deepCopyJSONValue\` recursively copies \`map[string]any\` and \`[]any\` so the entire JSON-decoded sub-tree is independent. Regression test \`family deep-copy survives mutation of nested object properties\` mutates the exposed copy's nested \`properties\`/\`required\` and asserts the cached schema is unchanged.

### 2. \`tools/list\` ordering was non-deterministic

\`assembleExposedTools\` iterated the \`familyBuckets\` Go map for emission, and consumed \`contributions\` in the order \`GetAllTools\` / \`GetAllToolsForSession\` appended them (driven by the registry's \`r.servers\` map iteration). Both sources are randomized. Repeated \`tools/list\` calls returned the same set in shifting orders — spurious \`tools/list_changed\` diffs downstream and flaky test fixtures upstream.

Sort \`contributions\` by \`serverName\` before bucketing; sort \`familyKeys\` by \`(family, toolName)\` before emission. Regression test \`tools/list order is stable across calls\` exercises this with mixed family + solo servers.

### 3. \`family.instanceArg\` collision with a tool input property

When \`family.instanceArg\` matched a name already declared in the tool's own \`InputSchema.Properties\`, the family-grouped emission silently overwrote the operator-defined property with the instance-selector enum. The original property's description, type, and constraints were lost.

The aggregator now detects the collision at the per-tool gate (same gate as description divergence) and falls back to per-server prefixing for that specific tool. Pattern mirrors the existing description-divergence handling; the two per-tool fallback emission paths share one helper (\`perServerTools\`).

\`Register\` is unchanged — collisions are detected during tool-list assembly, not at admission. This matches the codebase's existing convention of graceful degradation at the aggregator layer rather than admission-time rejection (see #698 for the fail-fast design discussion).

## Test framework changes

The collision fix needed two small framework changes so a BDD scenario could exercise it end-to-end:

- \`internal/testing/mock/server.go\` now publishes the operator-declared \`input_schema\` via \`mcp.NewToolWithRawSchema\`. Previously the mock fabricated an empty schema, so the aggregator's collision check was unreachable from BDD scenarios. Solo and dynamic tool registration paths both use the new \`toolWithSchema\` helper.
- \`internal/testing/muster_manager.go\`'s expected-tools pre-compute now mirrors the new collision logic so readiness checks wait for the correct exposed name (\`x_<server>_<tool>\`) when \`family.instanceArg\` collides.

## Tests

- New \`TestServerRegistry_FamilyGrouping\` subtests:
  - \`family deep-copy survives mutation of nested object properties\`
  - \`tools/list order is stable across calls\`
  - \`family.instanceArg colliding with a tool input property falls back to per-server prefixing\`
  - \`concurrent GetAllToolsForSession does not race\` (verifies #670's upsert semantics under \`go test -race\`)
- New scenario \`internal/testing/scenarios/mcpserver-family-instance-arg-name-collision.yaml\` exercises the collision fallback end-to-end.
- Full \`./muster test --concept mcpserver\` suite (68 scenarios) is green.

## Not included

- Fail-fast / reconciler-level conflict detection — see #698. Muster has no existing cross-CR domain-conflict pattern, and introducing one is larger scope than a #670 follow-up should carry.
- \`MCPServerInfo.Conditions\` exposure — useful only with a reconciler-set condition, blocked on #698.

Closes #697.


QuentinBisson (commented):

Bundling three fixes is justified — they share the call site (`assembleExposedTools`/`injectInstanceEnum`) and the `mock/server.go` harness change required to exercise (1) and (3) end-to-end. Don't split retroactively.

**Caveat on diff size**
- GitHub diff shows +3344/-471 but the actual fix commit (`f7a7f3e`) is ~+358/-20 across 5 files. The PR is "stacked on #670" but targets `main`, so #670's diff is inlined. Either re-target onto #670's branch or wait for #670 to merge — reviewers shouldn't re-review the whole family feature.

**Non-blocking — collision detection asymmetry**
- `assembleExposedTools` collision detection inspects only `entry.tools[0]` for the `entry.instanceArg` property. If server A's tool has no `cluster` property but server B's does, with otherwise-matching descriptions, the asymmetric collision is missed. Either union the `Properties` keys across `entry.tools[*]` or check all of them. Same symmetry the description-divergence check should have.

**Non-blocking — don't reinvent the wheel**
- `deepCopyJSONValue` reproduces `runtime.DeepCopyJSON` from `k8s.io/apimachinery/pkg/runtime`, which is already on your dep graph. One import swap drops ~20 hand-rolled lines plus the dev-only "what if a non-JSON value sneaks in" worry.
- `deepCopyJSONValue` doc claims "scalars are immutable in Go's value semantics" — true for the JSON-decoded subset (number/string/bool/nil), but `[]string`/`map[string]string` would silently fall through `default` and stay aliased. Tighten the doc to "JSON-decoded trees only contain `map[string]any` / `[]any` / scalars" or panic on other types in dev builds.

**Suggestion**
- CHANGELOG: missing `### Fixed` entries for all three fixes. These are observable behaviors (cache corruption from shallow copy, list-order churn breaking downstream diffs, silent property overwrite on instanceArg collision) and the consumer should see them.
- `sort.Slice` → `slices.SortFunc` for consistency with other call sites.

**Tests**
- Regression coverage for all three fixes is present:
  - Deep-copy: `family deep-copy survives mutation of nested object properties` — mutates nested `properties`/`required`, asserts cache untouched.
  - Stable ordering: `tools/list order is stable across calls` — 10 iterations, mixed family + solo.
  - Collision fallback: unit test + BDD scenario `mcpserver-family-instance-arg-name-collision.yaml`.
- Bonus: `concurrent GetAllToolsForSession does not race` adds race-detector coverage for #670.

Coverage shape is the strongest of the batch.

paurosello (approved):
Verified the three fixes:

- **Deep-copy**: `injectInstanceEnum` now uses `runtime.DeepCopyJSON` for the whole properties map (recursive into nested object properties + array items). Test `family deep-copy survives mutation of nested object properties` mutates the exposed copy's nested `filter.properties.namespace` and `filter.required`, asserts the cache is untouched.
- **Stable order**: contributions sorted by `serverName`; `familyKeys` sorted by `(family, toolName)` before emission. Test runs `GetAllTools` 10 times and asserts identical ordering.
- **Collision fallback**: `instanceArgCollides` now iterates **all** contributors (not `tools[0]` only). The asymmetric case where only one server declares the colliding property is covered by a dedicated test that asserts both `x_k8s-a_list_pods` and `x_k8s-b_list_pods` surface and the family-prefixed name does not.

`perServerTools` extracts the divergent-description + collision-fallback duplication cleanly. Mock server now propagates input_schema from YAML so BDD scenarios can exercise the collision path.

CI green, mergeable, CHANGELOG entries present.

PR #706, opened 2026-05-18
refactor: extract constants, drop ignore-string-values, fix gosec source

QuentinBisson:
## Summary

Follow-up to #700. Four logically distinct commits, kept separate for review:

1. **\`refactor(api): introduce ArgType enum for ArgMetadata.Type\`** — \`api.ArgMetadata.Type\` was a stringly-typed field with the valid values (\`string\`, \`number\`, \`integer\`, \`boolean\`, \`object\`, \`array\`) documented only in a comment. Introduces a typed \`api.ArgType\` enum and converts every adapter that produces \`ArgMetadata\` (workflow, mcpserver, orchestrator, events, app, metatools). \`validateType\` switches on the typed constants. Wire format unchanged.

2. **\`refactor(cli,api): share resource-type and JSON-key constants\`** — Lifts resource-type identifiers from \`cmd/mcp_types.go\` to \`internal/api/resource_types.go\` so \`internal/cli/table_formatter.go\` references one source of truth. Adds \`internal/api/json_keys.go\` for JSON Schema spec keys and muster API response field names, plus \`internal/cli/columns.go\` for kubectl-style table header labels. Sweeps the CLI table package consistently.

3. **\`refactor: extract JSON Schema, OAuth wire, and workflow op constants\`** — Eliminates the remaining duplicated string literals so the \`.golangci.yml\` \`ignore-string-values\` workaround can go away:
   - \`pkg/oauth/wire.go\` (RFC 6749/6750/7636/8693 vocabulary, adopted across aggregator + mock OAuth server)
   - \`internal/workflow/operations.go\` (\`opExecute\`, \`opValidate\`)
   - \`internal/agent/commands/constants.go\` (\`boolStringTrue\`, \`boolStringFalse\`)
   - JSON-key sweep across workflow, mcpserver, metatools, testing
   - Source-level gosec fixes: \`url.PathEscape\` for admin handler redirect (G710), \`context.Background()\` for SSO bootstrap goroutine (narrow G118 nolint with rationale), \`reflect.Pointer\` rename
   - \`.golangci.yml\`: drop the entire \`ignore-string-values\` block plus the temporary G118/G710/govet path exclusions from #700; keep only the test-file exclusion, the events/types.go G101 exclusion, and add narrow exclusions for the \`pkg/oauth/wire.go\` G101 false positive and the BDD mock OAuth server G710.

4. **\`refactor(config): share DefaultValkeyKeyPrefix across aggregator and oauth\`** — The \`muster:\` prefix was duplicated as a local default in four Valkey-backed stores. Lifts it to \`internal/config.DefaultValkeyKeyPrefix\` alongside the existing \`ValkeyConfig.KeyPrefix\` field. \`internal/aggregator/constants.go\` now holds only \`httpReadHeaderTimeout\`.

## Verification
- \`golangci-lint run -E=gosec -E=goconst -E=govet\` reports 0 issues with no \`ignore-string-values\` escape hatch.
- Full test suite passes.

## References
- #700


QuentinBisson (commented):

**Blocking — accidental commits**
- `.claude/worktrees/cleanup-serviceclass-doc-refs` and `.claude/worktrees/remove-teleport-support` are checked in as **gitlink submodule entries**. Almost certainly an unintended `git add` of a local-only worktree dir. Drop them before merge.

**Blocking — half a dedup**
- `internal/testing/mock/oauth_server.go` still hardcodes wire strings the new `pkg/oauth/wire.go` covers:
  - `r.FormValue("code")` (~line 1600)
  - `case "urn:ietf:params:oauth:grant-type:token-exchange"` (~line 1582) — `GrantTypeTokenExchange` constant exists
  - `"redirect_uri"`/`"scope"`/`"state"` query params (~lines 1540–1542)
  - `"error":` literal in `tokenExchangeError` (~line 1683)

  These survive only because the file got a blanket `G710` exclusion. The whole point of dropping `ignore-string-values` is to delete that escape hatch. Swap them, then narrow or drop the exclusion.

**Non-blocking — YAGNI**
- `pkg/oauth/wire.go` declares `HeaderContentType = "Content-Type"` with **zero call sites** in this PR. Eight `w.Header().Set("Content-Type", "application/json")` sites still hand-roll. Use at all 8 or delete the constant. Same audit for `FormFieldRequestedAud` and `FormFieldSubjectTokenT` — verify call sites.

**Non-blocking — DDD ownership**
- `cmd/mcp_types.go` keeps a local `const` block that re-aliases `api.ResourceType*` "to keep switches terse". Both packages owning the same vocabulary muddles ownership. Either reference `api.ResourceTypeService` directly (~12 chars per call site) or drop the api import and keep `cmd`'s constants as the source of truth.

**Non-blocking — `boolStringTrue`/`boolStringFalse`**
- `internal/agent/commands/constants.go` is two call sites in one package. Borderline YAGNI; `strconv.FormatBool(b)` at the call sites would be just as clear.

**Suggestion — DefaultValkeyKeyPrefix home**
- The constant bounced from `internal/aggregator` (#700) to `internal/config` (#706). Neither is right — it's a Valkey concern; `pkg/store/valkey` or co-located with `ValkeyConfig` is the right home. Worth one more move.

paurosello (approved):
Stack-rebased on top of the updated #700. Conflict in `internal/mcpserver/api_adapter.go` (new `family` arg from #670 collided with the constant extraction here) resolved by keeping the constant style and converting the new `family` schema to use `api.SchemaKey*` / `api.ArgType*`. Verified `go build ./...`, `go test ./internal/aggregator/... ./internal/mcpserver/...` green, `make lint` 0 issues.

PR #707, opened 2026-05-18
feat(agentgateway): PR 6/16 — Wire agentgateway into MCPServerReconciler

QuentinBisson:
Step **6/16** of the agentgateway adapter series. Plan: `/home/quentin/.claude/plans/declarative-forging-lynx.md`.

## What this lands

Compiles every MCPServer reconcile through `agentgateway.NewConfig` and applies it via the mode-appropriate `agentgateway.Applier`:

```go
// Filesystem mode — single long-lived yaml.Applier wired at startup
applier, _ := yaml.NewApplier(dir)
r := reconciler.NewMCPServerReconcilerFilesystem(
    orchestratorAPI, mcpServerMgr, registryHandler, applier, applier, // applier doubles as Deleter
).WithStatusUpdater(musterClient, namespace)

// Cluster mode — k8s.NewApplier constructed per Reconcile with ownerRef bound,
// StatusUpdater + namespace folded into the constructor (panics on nil).
r := reconciler.NewMCPServerReconcilerCluster(
    orchestratorAPI, mcpServerMgr, registryHandler,
    musterClient, k8sapply.Config{…},
    musterClient, namespace,
)
```

Two named constructors. The reconciler holds either `yamlApplier` (filesystem) or `k8sClient + k8sApplierCfg` (cluster) and branches per Reconcile in `applierFor`. No factory abstraction — the asymmetry lives in one place, where the application layer already lives.

## Why constructor-bound ownerRef + per-Reconcile k8s.Applier

Earlier draft passed `ownerRef` through `Apply` to allow one long-lived Applier in both modes. Reverted: `ownerRef` is K8s vocabulary and the port should stay free of it. Yaml mode keeps its long-lived applier; K8s mode constructs per Reconcile so `ownerRef` lives where it belongs — on the K8s adapter's constructor — and emitted objects cascade-delete cleanly.

## OwnerRef cache + fail-fast

- Cluster-mode constructor panics if `k8sClient` or `StatusUpdater` is nil — the K8s Applier rejects empty UID, so silent fallback would either error every reconcile or write orphan objects.
- `resolveOwnerRef` propagates `GetMCPServer` errors and rejects empty UID so the reconciler requeues with backoff instead of falling back silently.
- `applierFor` caches the resolved `OwnerReference` per `NamespacedName` so periodic status-sync requeues skip the GET and Applier rebuild. `reconcileDelete` invalidates the cache entry.

## Pure mappers extracted

`infoToMCPServerSpec`, `mcpServerAuthFromAPI`, and `tokenExchangeFromAPI` moved to the new `internal/reconciler/agentgateway/translate` subpackage so the reconciler focuses on orchestration. Pure functions, table-driven tests live next to the code.

## Files

- `internal/reconciler/mcpserver_reconciler.go` — two constructors, `applyConfig`, `applierFor` (with OwnerRef cache), `resolveOwnerRef`, status condition helpers. `mcpServerAPIVersion` now derived from `musterv1alpha1.GroupVersion.String()`.
- `internal/reconciler/agentgateway/translate/` — new subpackage holding `InfoToMCPServerSpec` / `MCPServerAuthFromAPI` / `TokenExchangeFromAPI` plus table-driven tests.
- `internal/reconciler/agentgateway/deleter.go` — optional `Deleter` port. yaml.Applier satisfies it; k8s.Applier does not (cascade via OwnerReferences).
- `internal/app/services.go` — `buildMCPServerReconciler` picks the right constructor based on `cfg.MusterConfig.Kubernetes`. Filesystem mode requires a non-empty configPath.
- `internal/reconciler/mcpserver_reconciler_test.go` + `mcpserver_reconciler_cluster_test.go` — existing tests updated; new coverage for constructor panics, cache lifecycle, applier-mode branching, stdio-in-cluster condition, and Deleter error propagation.

## Behaviour

- `NewConfig` rejection → status surfaces the error, no requeue (deterministic).
- `Apply` returns `k8s.ErrStdioNotSupportedInCluster` → `NotSupportedInCluster` status condition set with reason `StdioInClusterMode`. No requeue beyond the periodic status sync. (Routing on `agentgateway.ErrUnsupportedTransport` lands in PR #708 — sentinel hoist follow-up.)
- Successful Apply → previously-set `NotSupportedInCluster` condition is removed (rather than flipped to False) so `kubectl describe` stays quiet on the happy path.
- Other `Apply` errors → status surfaces, requeue with backoff.
- `resolveOwnerRef` GET failure → status surfaces, requeue (previously: silent empty-UID OwnerReference).

## Tests + lint

`go build ./...`, `go test ./...`, `go test -race ./internal/reconciler/...`, `golangci-lint run ./internal/reconciler/...` — all clean.


QuentinBisson (commented):

**Blocking — per-Reconcile cost & race**
- `applierFor` resolves the ownerRef and constructs a new `k8s.Applier` on every Reconcile, including periodic status-sync requeues that don't actually need to apply anything. That adds an extra `GetMCPServer` call per cycle and races with `syncStatus` which fetches the same object. Cache the ownerRef on the resource once UID is known, or return early before fetch when nothing has changed.

**Blocking — silent UID fallback**
- `resolveOwnerRef` falls back to an empty `UID` when `StatusUpdater == nil` or `GetMCPServer` fails. The K8s Applier doc claims it "rejects empty UID/APIVersion/Kind". Either you get a confusing apply error every Reconcile in this misconfig, or you write orphan objects. Fail fast — panic in the constructor if cluster-mode + nil StatusUpdater, consistent with the existing `k8sClient == nil` guard.

**Non-blocking — SRP creep**
- `mcpserver_reconciler.go` now owns both controller orchestration AND API→CRD spec translation (`infoToMCPServerSpec`, `mcpServerAuthFromAPI`, `tokenExchangeFromAPI`). These are pure mappers — move to `internal/api` or a translator subpackage so the reconciler stays orchestration-only (GS convention: `internal/controller/` orchestrates, `pkg/` holds logic).

**Non-blocking**
- `defaultGatewayName = "agentgateway"` and `agentgatewayConfigSubdir = "agentgateway"` in `app/services.go` are hardcoded. Comment acknowledges "environments that diverge will need a flag" — fine for now; file a follow-up so it doesn't get lost.
- `applyConfig` returns `(ReconcileResult, bool)` where the bool means "stop". Tri-state-via-pair smell. Consider `ReconcileResult.Stop bool` or aligning with the `(result, error, terminal bool)` pattern used elsewhere in the file.
- `clearNotSupportedInClusterCondition` fetches+conditionally-updates on every successful apply. `meta.RemoveStatusCondition` correctly returns false when absent so `mutate` short-circuits the UPDATE — but the GET still runs. At periodic-requeue cadence across N MCPServers that's N pointless GETs per cycle.
- `mcpServerAPIVersion = "muster.giantswarm.io/v1alpha1"` hardcoded — derive from the scheme via `musterv1alpha1.GroupVersion.String()` to avoid drift on group rename.
- Inconsistent nil-guards: the cluster constructor panics on nil `k8sClient` but not on nil `mcpServerManager`/`serviceRegistry`/`orchestratorAPI`. Guard all or none.

**Tests** — material gaps
- `applyConfig` returning `ErrStdioNotSupportedInCluster` → `setNotSupportedInClusterCondition`: uncovered.
- `applierFor` mode branching (yaml vs k8s): uncovered.
- `resolveOwnerRef` silent-fallback path: uncovered — exactly the edge worth pinning.
- `infoToMCPServerSpec` / `mcpServerAuthFromAPI` / `tokenExchangeFromAPI` mappers: untested. Round-trip pure functions — cheapest tests in the file.
- `reconcileDelete` Deleter error path: untested.

PR #708, opened 2026-05-18
refactor(agentgateway): hoist unsupported-transport sentinel into the port

QuentinBisson:
Follow-up refactor on top of PR #707 (step 5/9) — no series slot of its own.

Plan: `/home/quentin/.claude/plans/declarative-forging-lynx.md`.

## What this lands

`agentgateway.ErrUnsupportedTransport` — a port-level sentinel for transports an Applier cannot represent. The K8s applier's existing `ErrStdioNotSupportedInCluster` now wraps it:

```go
// internal/reconciler/agentgateway/applier.go
var ErrUnsupportedTransport = errors.New("agentgateway: backend transport not supported by this applier")

// internal/reconciler/agentgateway/k8s/applier.go
var ErrStdioNotSupportedInCluster = fmt.Errorf(
    "k8s applier: stdio MCPServers are not supported in cluster mode: %w",
    agentgateway.ErrUnsupportedTransport,
)
```

The reconciler routes on the port-level sentinel:

```go
if errors.Is(err, agentgateway.ErrUnsupportedTransport) {
    r.setNotSupportedInClusterCondition(ctx, req.Name, req.Namespace, err)
    return ReconcileResult{RequeueAfter: DefaultStatusSyncInterval}, true
}
```

## Why

PR #707 had the reconciler `errors.Is`-ing against `k8s.ErrStdioNotSupportedInCluster` — a concrete adapter symbol leaking into the application layer. Hoisting the sentinel up to the port:

- Adapter-agnostic callers route on `agentgateway.ErrUnsupportedTransport` without coupling to the K8s adapter for that purpose.
- Operators reading logs and the K8s applier's own tests still see the precise adapter wording.
- Future adapters (yaml applier rejecting some variant, a CEL applier rejecting transformations, …) wrap the same port sentinel without the reconciler needing a per-adapter import.

The reconciler still imports `k8sapply` for cluster-mode construction (`k8s.NewApplier`); that's an inevitable coupling for the constructor, separate from the routing concern fixed here.

## Tests + lint

`go test -race ./internal/reconciler/...`, `golangci-lint run ./internal/reconciler/...` — all clean.


QuentinBisson (commented):

Solid DDD cleanup — the reconciler stops reaching into `k8sapply` for routing, and the port owns the unsupported-transport contract. `fmt.Errorf("…: %w", agentgateway.ErrUnsupportedTransport)` is the correct Go 1.13+ wrap; `errors.Is` traversal works.

The `k8sapply` import in `mcpserver_reconciler.go` is still required for `k8sapply.Config` / `k8sapply.NewApplier` (cluster-mode construction). PR body acknowledges this — fine.

**Non-blocking — log line abstraction leak**
- `applyConfig`'s log line still says "MCPServer %s uses stdio; cluster mode does not support it yet" — but the routing is now generic. If a future adapter wraps `ErrUnsupportedTransport` for a non-stdio reason, the log will lie. Either gate the stdio-specific message on `errors.Is(err, k8sapply.ErrStdioNotSupportedInCluster)` and fall back to a generic message otherwise, or generalise to `err.Error()`. Minor, but the abstraction is leaking back through the log.

**Tests**
- No new test added for the sentinel wrap. Trivial assertion worth adding to `k8s/applier_test.go`: `require.ErrorIs(t, err, agentgateway.ErrUnsupportedTransport)` alongside whatever existing assertion on `ErrStdioNotSupportedInCluster`. Locks the wrap so a future refactor can't silently drop it.
- Reconciler-side test for `errors.Is(err, agentgateway.ErrUnsupportedTransport)` routing is the same gap called out in #707 — fixing it there covers this PR too.

PR #709, opened 2026-05-18
feat(agentgateway/subprocess): PR 7/16 — standalone process supervisor

QuentinBisson:
## What

Step **6/9** of the translator + agentgateway-as-subprocess series — see [`declarative-forging-lynx.md`](../../../.claude/plans/declarative-forging-lynx.md) §PR 7.

A standalone supervisor for one external process, used by filesystem-mode muster to run the agentgateway binary. **No MCPServer / muster knowledge** — this package has zero domain coupling.

`internal/agentgateway/subprocess/`:

- `Manager.Start(ctx, binary, args, env, readyProbe)` — spawn, pipe stdout/stderr into a `slog.Logger`, block on the caller-supplied readiness probe up to `WithStartupTimeout`.
- `Manager.Stop(ctx)` — SIGTERM the process group, wait up to `WithDrainTimeout`, SIGKILL on timeout. Idempotent.
- Restart-on-crash with capped exponential backoff (`WithBackoff(initial, max)`, `WithMaxRestarts(n)`).
- `HTTPReadyProbe(url, interval)` helper for the standard `:15021/healthz/ready` endpoint.

## Why this is its own package

- The supervisor knows nothing about MCPServer, agentgateway config, or muster's service locator. PR 9 wires it into `cmd/serve.go`.
- The existing in-repo process plumbing isn't reusable:
  - `internal/testing/process_unix.go` lives inside the BDD test harness — production code can't import it without a layering violation.
  - `internal/services/mcpserver/service.go` has its own backoff but is **poll-based** (orchestrator polls `nextRetryAfter`); the subprocess manager needs a **push-based** event loop.
  - `internal/mcpserver/client_stdio.go` defers process management to `mark3labs/mcp-go/client`, which is the wrong abstraction for managing the gateway itself.
- DRY of the ~30 lines of `Setpgid` / `Kill(-pid, sig)` would need a new shared package with two callers; YAGNI for this PR.

## Lifecycle correctness

Hot path is asserted by Go tests using the test binary itself as a fake agentgateway (`TestMain` dispatches on `SUBPROCESS_FAKE_MODE`):

| Concern | Test |
|---|---|
| Start blocks until ready, Stop reaps cleanly | `TestManager_Start_Stop_HappyPath` |
| Missing binary → error surfaces to Start | `TestManager_Start_BinaryMissing` |
| Stop before Start is a no-op; Stop is terminal | `TestManager_Stop_BeforeStart` |
| Stop is idempotent | `TestManager_Stop_Idempotent` |
| Concurrent Start: exactly one winner, rest get `ErrAlreadyRunning` | `TestManager_Start_Concurrent` |
| Crash → restart respects capped exponential backoff | `TestManager_Restart_OnCrash_WithBackoff` |
| **Stop racing the supervisor's first-spawn assignment still reaps** | `TestManager_StopDuringStartup` |
| **SIGTERM reaches grandchildren (whole-group reap)** | `TestManager_StopReapsGroup` |
| SIGTERM ignored → drain timeout → SIGKILL escalation | `TestManager_Stop_SIGKILLFallback` |
| stdout/stderr forwarded to the slog handler | `TestManager_CapturesStdoutStderr` |
| `New` rejects bad options (table-driven) | `TestNew_RejectsBadOptions` |
| HTTP readiness probe — 200, retry-until-ready, ctx cancel | `TestHTTPReadyProbe_*` |
| `signalProcessGroup` rejects pgid≤0 and treats ESRCH as success | `TestSignalProcessGroup_*` |
| **Goroutine leaks** across the suite | `goleak.VerifyTestMain` in `TestMain` |

No `time.Sleep` in tests; all waits go through `require.Eventually`.

### Stop-during-startup race

A subtle correctness fix landed in this PR after first review:

If `Stop()` fired between `cmd.Start()` returning and the supervisor's `m.pgid` assignment, `signalProcessGroup` would dispatch against `pgid=0` — SIGTERM was lost and the child leaked. Fixed by introducing a `spawned chan struct{}` that closes once the supervisor has either recorded the first successful `pgid` *or* given up before any spawn. `shutdown` waits on it (with caller-`ctx.Done()` as a parallel select) before reading `m.pgid`. The supervisor additionally self-signals SIGTERM under the lock if it observes `ctx.Err() != nil` immediately after recording `pgid`, covering restart-iteration races.

### Whole-group reap guarantee

`signalProcessGroup` used to fall back to `kill(pid, sig)` when `kill(-pgid, sig)` failed for non-`ESRCH` reasons — silently defeating the package's "any children the process spawned terminate with it" promise. The fallback is gone; non-`ESRCH` errors propagate. A new `TestManager_StopReapsGroup` spawns a fake binary that forks `/bin/sleep 300`, writes the grandchild PID to a file, and asserts the grandchild is gone after `Stop`.

### Readiness probe timeout

`HTTPReadyProbe` previously set `client.Timeout = interval`, so a slow first response was cancelled and immediately re-issued. Per-request timeout is now `2 * interval` while the poll cadence stays at `interval`.

## LOC

| File | LOC |
|---|---|
| `manager.go` | 412 |
| `options.go` | 79 |
| `ready.go` | 52 |
| `process_unix.go` | 33 |
| `process_windows.go` | 30 |
| `doc.go` | 17 |
| **prod total** | **623** |
| `manager_test.go` | 369 |
| `fake_binary_test.go` | 199 |
| `helpers_test.go` | 119 |
| `ready_test.go` | 56 |
| `process_unix_test.go` | 33 |
| **test total** | **776** |

## Verification

```
$ go test -race -count=1 -timeout 60s ./internal/agentgateway/subprocess/...
ok      github.com/giantswarm/muster/internal/agentgateway/subprocess   8.5s

$ go vet ./internal/agentgateway/subprocess/...
(clean)
```


QuentinBisson (commented):

DDD placement is correct — `internal/agentgateway/subprocess` is an infra adapter with no domain imports.

**Blocking — Stop-during-startup race**
- If `Stop` is called between `cmd.Start()` and the assignment of `m.cmd` / `m.pgid`, `pgid` is still `0` and `signalProcessGroup` sends to PID `0` — **SIGTERM is never sent to the child**, which then runs until parent exit. Capture `pgid` atomically inside the lock immediately after `cmd.Start`, and have `shutdown` wait on a "spawned" channel rather than reading `m.pgid` once.

**Blocking — silent group-fallback**
- `signal_unix.signalProcessGroup` falls back to `kill(pid, sig)` if the group send fails non-ESRCH. If `Setpgid: true` succeeded, the only reasons `kill(-pid, sig)` fails non-ESRCH are EPERM or a real race — falling back to `kill(pid)` defeats the whole "reap the whole group" guarantee silently. Log at minimum; consider returning the error instead.

**Non-blocking — YAGNI**
- `WithMaxRestarts`, `WithBackoff(initial, max)`, `Reload`/SIGHUP. The PR body literally says "agentgateway watches its config dir natively, so Reload is belt-and-suspenders". Delete `Reload` and its tests; ship in PR 9 if actually needed.

**Non-blocking — don't reinvent**
- `cenkalti/backoff/v4` is battle-tested. `waitBackoff` + the doubling math is fine but `backoff.NewExponentialBackOff` + `backoff.WithMaxRetries` shrinks `manager.go` and removes the `min(*backoff*2, max)` arithmetic.

**Non-blocking — misc**
- `Stop` after `Start` failure path calls `shutdown(context.Background())` — bypasses caller cancellation. Use `ctx`.
- `Manager` does lifecycle + log pumping + backoff + readiness orchestration. Acceptable at ~400 LOC, but `pumpLines` could move to its own file alongside `ready.go`.
- Windows path silently downgrades non-SIGKILL to `os.Interrupt`. Document that SIGHUP/Reload is a no-op on Windows, or return an error.
- `HTTPReadyProbe` uses `client.Timeout = interval`. Slow first response gets cancelled and immediately re-issued — wasted work. Separate per-request timeout (e.g. `2*interval`) is healthier.
- `envToSlice` sorts keys for determinism — fine, but document it's for test stability, not anything semantic about env ordering.

**Tests**
- Excellent shape: fake-binary via `TestMain` dispatch, `goleak.VerifyTestMain`, no `time.Sleep`, race detector run in CI. 83% on a subprocess package is strong.
- **Missing — the headline guarantee is unverified.** `doc.go` claims "any children the process spawned terminate with it". No test exercises that. Spawn the fake binary which spawns a grandchild (e.g., a long-`sleep` subshell), send SIGTERM, assert the grandchild PID also dies. Without this, the process-group plumbing could regress silently.
- `TestManager_Stop_DuringStartup` only asserts Start returns an error — should also assert `!pidExists(pid)`.

PR #710, opened 2026-05-19
feat(agentgateway/binary): PR 8/16 — pinned-release resolver

QuentinBisson:
Step **7/9** of the translator + agw-as-subprocess series (plan: `/home/quentin/.claude/plans/declarative-forging-lynx.md`).

Adds `internal/agentgateway/binary` — resolves the pinned `agentgateway` binary for filesystem mode via env var → local cache → atomic download from GitHub releases with **in-source pinned SHA-256 verification**.

## Trust model

The downloaded binary is verified against a hex-encoded SHA-256 hardcoded in [`checksums.go`](internal/agentgateway/binary/checksums.go), keyed by `<asset>/<version>`. This is the only trust anchor — the resolver does **not** fetch `<asset>.sha256` from the release, because that would be served by the same host as the binary, so a MITM with matching binary + companion file would defeat the check and TLS-pinning github.com already gives us the binary verification we'd get from a co-located digest.

Unknown `<os>/<arch>` at the pinned version returns `ErrUnpinnedPlatform` — fail-closed; adding a new platform requires a commit that explicitly inserts its hash. The download base URL is host-allowlisted to `github.com` (plus loopback for tests); arbitrary `WithDownloadBaseURL` URLs return `ErrUntrustedDownloadHost`.

A `make verify-checksums` target downloads upstream `<asset>.sha256` files for the current `PinnedVersion` and diffs them against the constants; this is the manual operator step when bumping `PinnedVersion`.

## Resolution order

1. `MUSTER_AGW_BINARY` env var. If set, the named file is used directly; broken paths error rather than fall through.
2. `<BaseDir>/agentgateway-v1.2.1[.exe]` cache (default `~/.config/muster/bin/`, overridable via `WithBaseDir`).
3. Download from `https://github.com/agentgateway/agentgateway/releases/download/v1.2.1/<asset>`, verify SHA-256 against the in-source constant, install atomically (temp file + chmod 0o755 + fsync + rename).

Network step is suppressed by `WithNoDownload`; cache miss in that mode returns `ErrBinaryNotFound`. Wired into `cmd/serve.go` in PR 8/9 (next).

## Verified release assets (v1.2.1)

| GOOS | GOARCH | Asset | Pinned SHA-256 (prefix) |
|---|---|---|---|
| linux | amd64 | `agentgateway-linux-amd64` | `90f549c7…3c3a` |
| linux | arm64 | `agentgateway-linux-arm64` | `bbdf5ac3…1a82` |
| darwin | arm64 | `agentgateway-darwin-arm64` | `e785ea8b…605f` |
| windows | amd64 | `agentgateway-windows-amd64.exe` | `6e517cf2…702f` |

Sources: `https://github.com/agentgateway/agentgateway/releases/download/v1.2.1/<asset>.sha256`. Cross-verified by `sha256sum` over each downloaded binary.

Unsupported (no upstream asset): `darwin/amd64`, `linux/arm`, `linux/ppc64le`, etc. → `ErrUnsupportedPlatform`. Supported `<os>/<arch>` with no pinned hash at the current version → `ErrUnpinnedPlatform`.

## `parseDigest` tightening

The function still exists (no production caller after this PR — kept as a defensive utility for future cases where we re-introduce companion-file verification, e.g. behind cosign attestations). It now requires a single non-empty line in canonical `sha256sum` format: a 64-char lowercase hex digest followed by a filename column that basename-matches `expectedFilename` (leading `*` stripped). Multi-line input is rejected — callers must pre-select the correct line.

## Why stdlib instead of `creativeprojects/go-selfupdate` (already a muster dep)

| Blocker | Source |
|---|---|
| `ChecksumValidator` does strict equality on the filename column | `validate.go:192-217` |
| `update.Apply` requires a pre-existing target file (first install fails) | `update/apply.go:128` |
| `decompressAndUpdate` expects an archive — agentgateway ships bare binaries | `update.go:93-105` |

Bypassing all three leaves only `NewGitHubSource + DetectVersion` — a thin GitHub-API wrapper that doesn't save meaningful code.

## LOC

| File | Prod | Tests |
|---|---|---|
| `doc.go` | 19 | — |
| `platform.go` | 39 | — |
| `options.go` | 82 | — |
| `resolver.go` | 249 | — |
| `checksums.go` | 12 | — |
| `platform_test.go` | — | 49 |
| `resolver_test.go` | — | 206 |
| `download_test.go` | — | 274 |
| `testutil_test.go` | — | 55 |
| **Total** | **401** | **584** |

## Verification

```
$ go test -race -count=1 ./internal/agentgateway/binary/...
ok  github.com/giantswarm/muster/internal/agentgateway/binary  1.2s

$ golangci-lint run ./internal/agentgateway/binary/...
0 issues.

$ make verify-checksums
agentgateway-linux-amd64        OK
agentgateway-linux-arm64        OK
agentgateway-darwin-arm64       OK
agentgateway-windows-amd64.exe  OK
```

Tests cover: env-var precedence (happy/broken/non-exec), cache hit / non-exec re-download / `NoDownload` error, download happy path / pinned-mismatch rejection / constant verification / context cancellation (no `.tmp.*` leak), BaseDir read-only-parent fault injection, HTTP 404 / 500, concurrent resolve (n=8) with leftover-tempfile assertion, host allowlist (subdomain attacks rejected), `parseDigest` single-line / multi-line / wrong-filename / hex-only / star-prefix, all platforms in the pinned map.

## Out of scope

- `--no-download` / `--agw-binary-dir` CLI flags — land in PR 8/9 (`cmd/serve.go` wiring).
- Cosign / sigstore signature verification — upstream doesn't publish signatures for v1.2.1. Flagged with a `TODO(security)` in `resolver.go`; once available we lift the host allowlist and verify per-asset attestations instead of (or alongside) the in-source constants.

Stacks on `feat/translator-08-subprocess` (#709). Once that merges, this PR's base auto-redirects.


QuentinBisson (commented):

**Blocking — security: checksum verification is effectively a no-op**

The checksum is fetched over the same channel as the binary, from the same host, with no out-of-band trust anchor. A MITM (corporate proxy, compromised CDN, registrar takeover) serves matching `<asset>` + `<asset>.sha256` and you install whatever they want. Fetching the digest alongside the artifact under plain TLS provides zero meaningful integrity beyond what TLS itself does for the binary alone.

Two options:

(a) **Embed the expected SHA256 for each `(os, arch, version)` tuple in the Go source** at the `PinnedVersion` bump site, and verify against the constant. The fetched `.sha256` becomes redundant and can be dropped. Renovate or a make target can re-pin on version bumps.

(b) **Require cosign signature verification.** If upstream agentgateway doesn't sign v1.2.1 yet (the PR body suggests this), push them to sign and gate adoption on it.

This is the actual security boundary of the resolver, and right now it is doing no useful work.

**Blocking — base URL override**
- `defaultDownloadBaseURL = "https://github.com/..."` is mutable via `WithDownloadBaseURL`. Fine for tests, but the env-var override (or any future config knob) could redirect production downloads. Pin the host (`https://github.com/agentgateway/`) at the resolver level and either restrict `WithDownloadBaseURL` to a test-only build tag or reject non-`github.com` hosts in production.

**Blocking-ish — parseDigest format coupling**
- `parseDigest` accepts the first 64-char hex token. Upstream serves a single-line `<asset>.sha256` today, so it works. If upstream ever switches to a combined `SHA256SUMS` file (common pattern), the resolver picks the first hash, not the one for the requested asset. Match the asset name; add a test that a multi-line file is rejected unless the matching filename is found.

**Non-blocking**
- `http.Client.Timeout: 5 * time.Minute` is a wall-clock timeout for header + body. Slow link + 50MB binary on a poor connection trips this. Use `Transport.ResponseHeaderTimeout` + a context-aware `io.Copy`, or expose `WithDownloadTimeout`.
- `checkExecutable` on Windows skips the exec-bit check — fine — but a cached `.exe` truncated to zero bytes passes. Add a `Size() > 0` check.
- The `outputs/` prefix workaround should be raised upstream — agentgateway's build artifact leak isn't yours to permanently absorb.
- `hashicorp/go-getter` rejection was right (drags AWS/GCS). `creativeprojects/go-selfupdate` rejection was reasoned. Plain stdlib is correct.

**Tests**
- `TestResolve_ChecksumMismatch` is the most important test in the package and it's present — good. Once the embedded-SHA option above lands, add a test that a `.sha256` from the network is *not consulted* when the constant exists.
- `TestResolve_Concurrent` with `n=2` doesn't reliably exercise the `os.Rename` race (both may finish download before either renames). Bump to `n=8`; assert no leftover `.tmp.*` files and that only one rename "won" (e.g. inode/mtime stable across resolvers).
- No test that `MUSTER_AGW_BINARY` pointing at a *directory* is rejected — `checkExecutable` does check `IsDir`, but exercise it.

PR #720, opened 2026-05-19
feat(aggregator): PR 11+15a — upstream-proxy rewrite + spec.suspended landing early

QuentinBisson:
Step **11/16** of the agentgateway adapter series. Plan: `/home/quentin/.claude/plans/declarative-forging-lynx.md`.

## Summary

muster's aggregator now dials agentgateway at `<UpstreamProxy>/mcp/<server-name>` (streamable-http, regardless of `spec.type`) for every external MCPServer. Federation registration is reconciler-driven via `aggregator.RegisterUpstream` / `DeregisterUpstream` instead of orchestrator-driven `processAutoStartMCPServers`. With the dial path unified through agentgateway, every component that lived only to support the orchestrator-spawned MCPServer service path is deleted in the same PR.

This is the unblocking change for the muster-in-front-of-agentgateway intermediate topology: muster keeps token exchange, server-family grouping (`spec.family`, #670), and ADR-006 session-scoped tool filtering; agentgateway becomes transparent L7 middleware for tracing/audit/metrics.

## Why this also ships PR 15's `spec.suspended`

`MCPServer.spec.suspended` is the persistence layer for `core_service_stop`. Once the orchestrator stops owning MCPServer lifecycle (this PR), the previous in-memory `userStopped` set on the aggregator has no home — operator intent would not survive a muster restart. Splitting the suspended flag out into PR 15 as originally scoped would have left `core_service_stop` / `_start` / `_restart` half-working against MCPServer names for one PR cycle. The CRD field, `SetSuspended`, the orchestrator-adapter Start/Stop/Restart branches, and the `Suspended` reconcile branch land together here.

## Configuration

`AggregatorConfig.UpstreamProxy` is **required** at startup; muster fails fast when unset.

- **Filesystem mode** (`!cfg.MusterConfig.Kubernetes`): defaults to `http://localhost:8080` (the agentgateway subprocess listener, gated on PR 9's `:15021/healthz/ready` probe).
- **Cluster mode** (`cfg.MusterConfig.Kubernetes`): reads `MUSTER_AGW_UPSTREAM_URL` from the muster pod's environment. The Helm chart sources it from `muster.agentgateway.upstreamURL`; when that value is unset the chart's `muster.agentgatewayUpstreamURL` helper falls back to `http://agentgateway.<release-namespace>.svc.cluster.local:8080`. Operators with a divergent agentgateway topology override `muster.agentgateway.upstreamURL` explicitly.

### Aggregator
- `AggregatorConfig.UpstreamProxy` field + `proxyURLFor` helper.
- Dial-site rewrites in `connection_helper.go` (`establishConnection`, `EstablishConnectionWithTokenForwarding`, `EstablishConnectionWithTokenExchange`): dial URL substituted; `serverInfo.URL` retained for display/audit.
- `AggregatorManager.RegisterUpstream` / `DeregisterUpstream` / `UpstreamServerState` plus interface entries on `api.AggregatorHandler`.
- `AggregatorManager.RegisterUpstream` handles the OAuth pending-auth path via `*mcpserver.AuthRequiredError` → `RegisterPendingAuthWithConfig` (display URL stays `info.URL`; dial URL is the proxy).

### Reconciler
- `applyConfig` success → `RegisterUpstream(name)` when `AutoStart=true`.
- `reconcileDelete` → `DeregisterUpstream(name)` + applier `Delete`.
- `applyStatusFromAggregator` replumbs status sync to read from `api.AggregatorHandler.UpstreamServerState` (Connected / AuthRequired / Absent).
- `spec.suspended=true` short-circuits reconcile to a deregister + status sync; operator intent persists on the CRD instead of in-memory aggregator state.

### Dead code deleted in this PR
- `internal/orchestrator/orchestrator.go`: `processAutoStartMCPServers`, `createMCPServerService`, `handleAuthRequiredServer`, `retryFailedMCPServers`, `attemptReconnectFailedServers`, `shouldAttemptRetry`, `waitForMCPServerRegistration`, `Config.DisableMCPServerAutoStart`. Net file shrinks from 527 → 232 LOC.
- `internal/reconciler/mcpserver_reconciler.go`: `WithDisableLocalSpawn` + `disableLocalSpawn` field + every branch. `orchestratorAPI` + `serviceRegistry` parameters dropped from the constructors.
- `internal/aggregator/manager.go`: `registerHealthyMCPServers`, `registerSingleServer`, `deregisterSingleServer`, `EventHandler` wiring, `retryFailedRegistrations`, `attemptPendingRegistrations`, `ManualRefresh`, `isServerAuthRequired`, `isServerSSOBased`, `GetEventHandler`. `internal/aggregator/event_handler.go` + `event_handler_test.go` gone.
- `internal/services/aggregator/` entire package. Its `AggregatorService` `BaseService` wrapper, `onManagerErrorCallback`, `Restart`, `GetServiceData`, `service_test.go`. `internal/aggregator/api_adapter.go` (new) wraps `*AggregatorManager` directly; `internal/app/services.go` and `runOrchestrator` construct/supervise the manager directly.

### Helm chart
- `helm/muster/templates/deployment.yaml`: `MUSTER_AGW_UPSTREAM_URL` env var alongside `K8S_*`.
- `helm/muster/templates/_helpers.tpl`: `muster.agentgatewayUpstreamURL` template with namespaced default.
- `helm/muster/values.yaml`: `muster.agentgateway.upstreamURL` (empty default → helper-resolved fallback).
- `helm/muster/values.schema.json` updated to match.

### Cluster-mode stdio MCPServers stay unsupported (deviates from plan)

The plan promised cluster-mode stdio would unblock here "as a side effect" of the proxy rewrite. It does not. Upstream `AgentgatewayBackend` CRD at `controller/api/v1alpha1/agentgateway/agentgateway_backend_types.go:474` has no stdio variant — `McpTarget` is `{Host, BackendRef, Port, Path, Protocol}` with `MCPProtocol ∈ {StreamableHTTP, SSE}` only. So `ErrStdioNotSupportedInCluster` + `ConditionTypeNotSupportedInCluster` (added in PR 3 / #692) stay in place; cluster-stdio remains a future-PR concern that needs upstream agentgateway support before muster can deliver it.

Either the plan needs to drop the side-effect claim, or a follow-up has to land cluster-stdio through per-MCPServer pod isolation. Leaving the symbol in.

## Net diff

55 files changed, +2,869 / -4,608. Larger than the original PR 11 envelope because (a) the orchestrator + `services/aggregator` excisions that were going to be PR 13 + PR 14(d) are bundled here — they are tightly coupled to the dial-path rewrite and splitting them would have required keeping dead code alive across two PRs — and (b) `spec.suspended` (originally PR 15) lands now for the reason above.

## Verification evidence

- `make test -race` clean across `internal/aggregator/...`, `internal/reconciler/...`, `internal/app/...`, `internal/orchestrator/...`.
- `make helm-test` clean (64 tests across 6 suites).
- `make helm-lint` clean.
- `go vet ./...` clean.
- New unit tests: `internal/aggregator/upstream_test.go` (proxyURLFor), `internal/reconciler/mcpserver_reconciler_test.go` (RegisterUpstream on AutoStart=true, skipped on AutoStart=false, NotSupportedInCluster on stdio sentinel, DeregisterUpstream on delete, register-failure requeue, AuthRequired mapping from aggregator state).

## Open items the reviewer should weigh in on

1. **`WWW-Authenticate` round-trip through agentgateway v1.2.1.** The OAuth pending-auth path relies on the 401 + `resource_metadata` reaching muster verbatim. The PR 16 (#724) BDD scenarios assert this; if agentgateway strips the header, we will need an upstream issue / workaround before merging.
2. **`MCPServer.status.consecutiveFailures` / `lastAttempt` / `nextRetryAfter`** become permanently zero (the per-service retry state machine is gone). Surfaced in the CHANGELOG; if a Grafana dashboard or downstream automation depends on these fields, that consumer needs a migration plan.
3. **`core_service_start/stop/restart` adapter fall-through** at `internal/orchestrator/api_adapter.go:62-110` silently returns to `a.orchestrator.StartService(name)` when the MCPServer manager lookup fails, masking the "service not found" error path. Worth a follow-up to surface `api.ErrServiceNotFound` explicitly.
4. **Cluster-mode stdio unblock claim.** See "Cluster-mode stdio MCPServers stay unsupported" — the plan needs reconciling against the CRD reality.

## Stacked PRs above this one (the remaining /16 plan)

- **PR 12 / #721** — Reconciler + aggregator shape simplification: collapse `Applier` + `Deleter` into one port, single reconciler constructor, lift `OwnerRefFor` to a free function.
- **PRs 13+14 / #722** — Excise orchestrator + services packages outright; post-pivot small cleanups (drop `subprocess.Manager.Reload`, `WithListenerName`, factory dispatch, SSE/stdio client packages).
- **PR 15 / #723** — Retire `core_service_{start,stop,restart}` against MCPServer names; rename `restart` → `core_mcpserver_reconnect`. (The `spec.suspended` field itself lands in this PR — see "Why this also ships PR 15's `spec.suspended`" above.)
- **PR 16 / #724** — Docs (filesystem-mode quickstart + cluster-mode CRD workflow + agentgateway deployment runbook) + regression / feature / end-to-end BDD scenarios.

## Branch base

Cut off `origin/feat/translator-pr10-federation` (the cumulative state of PRs 1-10 of the agentgateway-as-subprocess series); rebase to `main` once #714 (PR 9/16) and #717 (PR 10/16) merge.


QuentinBisson:
Stop-conditions and the userStopped / readiness majors from the review are fixed in #733 (draft, stacked on this branch).

- `spec.family` server-grouping restored end-to-end in the real aggregator (registry + dispatch + session-scoped variants); `mcpserver-family-*` BDD scenarios now exercise the real aggregator rather than the harness mock.
- `oauth-sso-state-sync-after-login.yaml` reverted to expect `connected`; root cause fixed at the source via `UpstreamServerStateForSession(ctx, name)` on the handler + session-aware `core_service_list` / `core_service_status`.
- `MCPServer.spec.suspended` replaces the in-memory `userStopped` set / `MarkUserStopped` / `MarkUserStarted` so operator intent survives restarts and concurrent reconciles.
- `initializeWithConnectRetry` replaced by a synchronous `:15021/healthz/ready` handshake + tight `ECONNREFUSED` retry covering the per-route bind race.

`muster test --concept mcpserver --parallel 4`: 65/65 (vs the 58/58 baseline that had the relaxation and family deletion masked).

PR #725, opened 2026-05-19
refactor(crds): split muster CRDs + helm chart hardening

QuentinBisson:
## Summary

Two related shifts to make the muster Helm chart usable as a sub-chart and safe under Giant Swarm production policies.

### 1. CRD lifecycle split

Helm 3 only special-cases the top-level chart's `crds/` directory — sub-chart `crds/` dirs are silently ignored. Bundling CRDs through `helm/muster/templates/crds.yaml` made the chart unusable as a sub-chart (e.g. inside `agentic-platform`) and leaked the CRD lifecycle into every `helm upgrade` of the control plane.

CRDs now ship in a sibling chart `muster-crds` whose only payload is the `crds/` directory. The muster chart no longer renders any CRDs.

- New chart `helm/muster-crds/` containing `mcpservers.muster.giantswarm.io` and `workflows.muster.giantswarm.io`.
- `Makefile.crd.mk`: `controller-gen` writes to `helm/muster-crds/crds/`.
- `.github/workflows/ci.yaml`: CRD-drift check follows the new path.
- `.circleci/config.yml`: parallel `architect/push-to-app-catalog` invocations build, test, and publish `muster-crds` alongside `muster` on the same release tags.
- `helm/muster/templates/crds.yaml` removed.
- `helm/muster/values.yaml`: `crds.install` defaults to `false`, deprecated, scheduled for removal in chart v0.2.0.
- `helm/muster-crds/UPGRADE.md` documents the schema-change ratchet (Kubernetes deprecation policy) and operator ordering with respect to the `muster` App CR.

Operators MUST install / upgrade `muster-crds` **before** `muster`. On the Giant Swarm App platform, encode this via `spec.dependsOn` on the muster App CR. On clusters with the CRDs already installed by an earlier muster release, Helm will refuse to take ownership unless the CRDs carry the matching release metadata — `UPGRADE.md` has the one-time adoption snippet.

### 2. Helm chart hardening (driven by the glean rollout)

- `CiliumNetworkPolicy`: unconditional egress to `app.kubernetes.io/name=agentgateway:8080` in the release namespace so muster's aggregator can dial the agentgateway data-plane on the upstream-proxy path. No-op on Cilium when the data-plane isn't deployed.
- `Deployment.spec.revisionHistoryLimit`: new value (default `3`) so frequent rollouts don't leak zero-replica ReplicaSets across the cluster.
- `resources.{requests,limits}.ephemeral-storage` (50Mi / 100Mi) on the muster container — `/tmp` is mounted as `emptyDir` and Kyverno's audit policy on GS workload clusters warns without an explicit limit; strict-enforce clusters reject the pod outright.

## Follow-up

Companion PR on `giantswarm/agentic-platform` (#1, open) drops the `agentgateway-crds` sub-chart dependency, switches to the same `dependsOn` pattern, and broadens the umbrella CNP with the same DNS / cluster-ingress / muster-aggregator allowances.


paurosello:
Why is this a problem? CRDs get upgraded when the app is installed, the only problem I could see is interchart dependency but then I would extract them to a different chart like "muster-crds" but never using the crds folder as that makes it imposible to update CRDs right?

```
Bundling CRDs through helm/muster/templates/crds.yaml made the chart unusable as a sub-chart (e.g. inside agentic-platform) and leaked the CRD lifecycle into every helm upgrade of the control plane.
```

QuentinBisson:
Closing — the stated rationale was wrong.

Pre-#725 muster shipped a `helm/muster/templates/crds.yaml` file-glob that loads `crds/*` via `$.Files.Get`. That template **does** render when muster is consumed as a sub-chart — Helm only restricts the `crds/` directory special-case to the top-level chart, not all CRD handling. So the claim "broke installation when the muster chart was pulled as a sub-chart" was false. The umbrella was already getting muster's CRDs through the template.

The split does have real benefits (uninstall doesn't cascade-delete `MCPServer` / `Workflow` CRs, CRD schema upgrades can ratchet on a different cadence than the data plane), but they don't outweigh the operational simplicity of a single chart for muster's current scale. We can revisit if either of those concerns bites in practice — `helm.sh/resource-policy: keep` on the templates/crds.yaml output would also address the cascade-delete risk without a full chart split.

Contrast with `agentgateway-crds`: that one is a real upstream split — `cr.agentgateway.dev/charts/agentgateway` ships only the controller, `cr.agentgateway.dev/charts/agentgateway-crds` ships only the CRDs. We have no choice but to consume both.

Will keep the chart-hardening pieces from this PR (CNP agentgateway egress, revisionHistoryLimit, ephemeral-storage limits) and reopen those as a separate PR.

QuentinBisson:
I think I and claude were confused with the agentgateway CRD :( sorry

PR #732, opened 2026-05-20
Update Chart.yaml metadata

marians:
Primarily, this PR should change ownership for the chart and alerts related to it to bumblebee

It also removes the generic keyword "platform-engineering".

### DIFF
--- a/helm/muster/Chart.yaml
+++ b/helm/muster/Chart.yaml
@@ -13,6 +13,9 @@ keywords:
   - ai-agents
   - model-context-protocol
   - aggregator
+maintainers:
+  - name: Giant Swarm
+    email: team-bumblebee@giantswarm.io
 annotations:
   io.giantswarm.application.audience: all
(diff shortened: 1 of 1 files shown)


helm/muster/Chart.yaml
```diff
@@ -12,11 +12,7 @@ keywords:
   - mcp
   - ai-agents
   - model-context-protocol
-  - platform-engineering
   - aggregator
-maintainers:
-  - name: Giant Swarm
-    email: team-planeteers@giantswarm.io
```
QuentinBisson, line 19: I think we should keep the maintainers but switch it to bumblebee right?

marians: I've rarely seen this in any GS chart. But I don't mind if you want to keep it.

marians: Added back

QuentinBisson (approved):
Thank you

PR #734, opened 2026-05-20
feat(helm): networkPolicy.flavor — cilium or vanilla Kubernetes

QuentinBisson:
## Summary

Adds `networkPolicy.flavor` to the muster chart so consumers running on non-Cilium clusters can render a vanilla `networking.k8s.io/v1` NetworkPolicy instead of a CiliumNetworkPolicy. Surfaces from the agentic-platform umbrella refactor — same shape as the umbrella's `networkPolicy.flavor` so a single flavor value can plumb through.

`ciliumNetworkPolicy.*` is removed (no legacy alias). Pre-1.0 chart, downstream consumers update their values once.

**Rebased on top of #735** so the agentgateway:8080 egress added there is preserved in the cilium template AND mirrored into the new vanilla NetworkPolicy template. Merge #735 first; this PR's diff against main becomes only the network-policy refactor.

## Migration

| Old | New |
|---|---|
| `ciliumNetworkPolicy.enabled: true` | `networkPolicy.enabled: true` (defaults to `flavor: cilium`) |
| `ciliumNetworkPolicy.allowClusterIngress: true` | `networkPolicy.cilium.allowClusterIngress: true` |
| `ciliumNetworkPolicy.labels` / `annotations` | `networkPolicy.labels` / `annotations` |

## kubernetes flavor

Vanilla `NetworkPolicy` has no entity selectors, no FQDN egress. Replacements:

| Cilium concept | Kubernetes flavor |
|---|---|
| `toEntities: [kube-apiserver]` | `ipBlock: {{ networkPolicy.kubernetes.apiServerCIDR }}` (default `0.0.0.0/0`) |
| `toEntities: [cluster]` (via `allowClusterIngress`) | `ipBlock: {{ networkPolicy.kubernetes.clusterCIDR }}` — empty disables |
| `toEntities: [world]` | `ipBlock: 0.0.0.0/0 except: {{ worldExcludedCIDRs }}` |
| `toFQDNs:` | not supported |

Both flavors include the agentgateway:8080 egress added in #735.

## Tests

`helm unittest helm/muster` — 78 passed. `tests/networkpolicy_test.yaml` covers:

- Neither template renders when disabled.
- `cilium` flavor renders CiliumNetworkPolicy and NOT vanilla NetworkPolicy.
- `kubernetes` flavor renders vanilla NetworkPolicy and NOT CiliumNetworkPolicy.
- `networkPolicy.cilium.allowClusterIngress: true` adds the cluster entity egress on 80/443.
[...]

PR #735, opened 2026-05-20
feat(helm): move CRDs to files/, configurable annotations, chart hardening

QuentinBisson:
## Summary

Two changes around the muster Helm chart, split from the closed PR #725.

### CRDs

CRD source files move from `helm/muster/crds/` to `helm/muster/files/crds/`. `files/` has no Helm 3 `crds/` special-case, so the rendering path (`helm/muster/templates/crds.yaml` loader) is identical whether muster is installed standalone or consumed as a sub-chart. Eliminates the dual-render ambiguity that `helm/muster/crds/` had at top-level install (Helm 3 would auto-install from `crds/` AND the loader template would render them too, with mismatched ownership).

The loader merges `crds.annotations` into each CRD's `metadata.annotations`. Default:

```yaml
crds:
  install: true
  annotations:
    helm.sh/resource-policy: keep
```

`helm.sh/resource-policy: keep` keeps the CRDs (and every `MCPServer` / `Workflow` CR that depends on them) around on `helm uninstall`. Operators can clear (`crds.annotations: {}`) or extend the map.

`Makefile.crd.mk`'s controller-gen target updates to write into the new path. `.github/workflows/ci.yaml`'s CRD drift check follows.

### Chart hardening (from #725)

- `Deployment.spec.revisionHistoryLimit` value (default `3`) so frequent rollouts don't leak zero-replica ReplicaSets.
- `resources.{requests,limits}.ephemeral-storage` (`50Mi` / `100Mi`) — `/tmp` is an emptyDir on the muster container; Kyverno's resource-limits policy on Giant Swarm workload clusters audits (or rejects) pods without an explicit ephemeral-storage value.
- `CiliumNetworkPolicy` egress to `app.kubernetes.io/name=agentgateway:8080` in the release namespace so muster can dial the agentgateway data-plane on the upstream-proxy path. No-op when agentgateway isn't deployed.

## Tests

`helm unittest helm/muster`: 68 passed. New `tests/crds_test.yaml` covers:
- `crds.install: false` → no documents rendered.
- Default render emits both CRDs with `helm.sh/resource-policy: keep`.
- Arbitrary annotations merge into `metadata.annotations` alongside the default.

## Downstream impact

The agentic-platform umbrella (PR giantswarm/agentic-platform#1) currently sets `muster.ciliumNetworkPolicy.enabled: true` and consumes muster's CRDs via the sub-chart's `templates/crds.yaml`. Both continue to work transparently — the umbrella picks up the new annotation behaviour without changes.


paurosello (approved):
LGTM. files/crds/ + fromYaml loader is the right approach (avoids Helm 3 crds/ dual-render). Rebase on top of #734 before merge — both touch values.yaml/schema/README/CNP.

PR #750, opened 2026-05-27
feat(helm): extract muster CRDs into a standalone muster-crds chart

paurosello:
## What

Extracts muster's CRDs (`MCPServer`, `Workflow`) into a new standalone chart `helm/muster-crds/`. The muster **app** chart no longer renders CRDs.

## Why

A downstream `agentic-platform-crds` umbrella (in `giantswarm/agentic-platform`) needs to bundle the agentgateway **and** muster CRDs as sub-chart dependencies, so operators install **one** CRDs release first and the workload releases after — Flux/Argo-agnostic ordering, with CRD and workload lifecycles decoupled.

> Note on the prior attempt (#725, closed): that PR's rationale was wrong — muster's `templates/crds.yaml` `$.Files.Glob` loader renders fine as a sub-chart; the real uninstall bug was always *agentgateway's* CRDs. It also used a top-level `crds/` folder (Helm 3 treats that as **install-only** — no upgrades). **This PR fixes both:** the rationale above is the real one (single downstream CRDs umbrella), and the CRDs keep shipping under `files/crds/` rendered via `templates/` (upgradable) with `helm.sh/resource-policy: keep`.

## Changes

- **New `helm/muster-crds/`** — `Chart.yaml`, `files/crds/{mcpservers,workflows}` (git-renamed from the app chart, content unchanged, `controller-gen` v0.21.0 preserved), `templates/crds.yaml` loader (no `install` guard — installing this chart *is* how you install the CRDs), keep-annotation default, schema/docs/tests.
- **muster app chart** — `templates/crds.yaml` removed; `crds.install` now defaults `false` and is a deprecated compatibility shim (kept so a downstream setting `muster.crds.install=false` still validates; removal scheduled next release).
- **CI/build** — `.circleci/config.yml` gains parallel `*-muster-crds-*` publish jobs (build, ATS tests, branch push, tag push) mirroring muster's shape, versioned in lockstep off the same tag; `make generate-crds` + the CI drift check repointed to `helm/muster-crds/files/crds/`.
- **Docs** — CHANGELOG entry; repointed 4 stale `helm/muster/crds/...` paths in `docs/operations/installation.md` (that path was already non-existent).

## Operator impact

Install/upgrade **`muster-crds` before `muster`**. Existing installs that relied on the muster app chart shipping CRDs must add the `muster-crds` release.

## Verification (local)

- `helm lint` passes for both charts.
- `helm template helm/muster-crds` → 2 CRDs, both `helm.sh/resource-policy: keep`.
- `helm template helm/muster` → **0** CRDs.
- `make generate-crds` (controller-gen) → no schema drift; v0.21.0 annotation preserved.
- `helm package helm/muster-crds` → ok.

## TODO / follow-up

- `helm/muster-crds/values.schema.json` is hand-authored (compact draft-07, matching the existing checked-in muster schema). A `devctl gen` pass may rewrite it into the canonical bundled draft-2020 form — same divergence as the current muster chart. Flagging for a devctl regen.
- `helm-test` / unittest CI currently runs only `helm/muster`; the moved `helm/muster-crds/tests/crds_test.yaml` should be added to that run.

Paired with the `giantswarm/agentic-platform` CRDs-split PR (link to follow).


QuentinBisson (approved):

paurosello:
Downstream consumer PR: giantswarm/agentic-platform#12 (adds `agentic-platform-crds`, which depends on this `muster-crds` chart). Its CI's dependency-build stays red until `muster-crds` is released to the prod catalog from this PR.

PR #757, opened 2026-05-28
chore: superseded — split into #772 and #773

QuentinBisson:
Superseded. This PR has been split into two focused PRs:

- #772 — AllowedClaims in TrustedIssuer, drop KubernetesSATrusts, fix JWT signing key wiring (land this first)
- #773 — machinePrincipals M2M token exchange (stacked on #772; interim design pending giantswarm/mcp-oauth#424)

The "session-aware aggregator" section in the original description was copy-paste noise — that work landed separately via #765.


paurosello:
Pushed a small follow-up commit (749f9b4) addressing review feedback directly on the branch — shout if you'd rather I'd left these as comments:

- **Dead code:** removed the now-orphaned `k8sSASubPattern` + its test (last production caller went away with `KubernetesSATrusts`) and the unused `strings` import.
- **PII:** the machine-principal exchange log now redacts the email via `truncateEmail`, matching the existing SSO logging convention.
- **Tests:** added handler-level tests for `machinePrincipalExchangeMiddleware.ServeHTTP` (fall-through routing) plus `writeExchangeError`/`writeJSONError`.

Verified: `go build ./...`, `go vet`, full `internal/server` suite, and `golangci-lint` (0 issues).

Two things left for you:
- The happy-path injection (`ExchangeSubjectToken`) and the `unsupported_grant_type` mapping aren't unit-tested — both need an interface/constructor seam that doesn't exist, so I didn't refactor for it. These look like they're meant to be covered by the **session-aware aggregator BDD scenarios mentioned in the PR description, which aren't actually in this diff** — can you confirm whether those got dropped?

PR #772, opened 2026-06-02
feat: AllowedClaims in TrustedIssuer, drop KubernetesSATrusts, fix JWT signing key wiring

QuentinBisson:
## What

Two related changes that unblock functional JWT mode, plus a new knob for private-IP OIDC discovery.

**AllowedClaims / KubernetesSATrusts cleanup (mcp-oauth v0.2.171+)**
- Add `AllowedClaims map[string]string` to `TrustedIssuerConfig`; wire it through `buildOAuthServerOptions`
- Remove `KubernetesSATrusts` / `K8sSATrustConfig` — now expressed via `trustedIssuers` with `allowedClaims`
- Fix dpop/valkey import path for mcp-oauth v0.2.170+
- Remove orphaned `k8sSASubPattern` and dead code

**JWT mode signing key wiring (previously broken)**
`Config.Validate()` hard-fails unless signing key fields are all set when JWT mode is on. These were never populated, so `enableJWTMode: true` always fell into degraded mode silently.

**`allowPrivateIPOIDC` — new Dex config knob (this PR)**
Clusters where the public Dex hostname resolves to an RFC 1918 address (e.g. Azure internal LB, air-gapped) get `context deadline exceeded` on OIDC discovery and start in degraded mode. `muster.oauth.server.dex.allowPrivateIPOIDC: true` disables the SSRF guard for that specific path, matching the existing pattern for JWKS (`allowPrivateIPJWKS`) and token exchange.

Requires: **mcp-oauth#427** (`dex.Config.AllowPrivateIP`)


paurosello (approved):
## ✅ Approved — all CI green

Reviewed the full diff, resolved the merge conflicts (merged `main`, kept `mcp-oauth v0.2.186` + `mcp-toolkit v0.2.5`), and added the two missing docs rows (`jwtSigningKeyFile` + `dex.allowPrivateIPOIDC`). All checks pass: GH Actions (Lint and Test, pre-commit, gitleaks, check-values-schema) and every CircleCI job.

### What this does
- **`AllowedClaims` / `KubernetesSATrusts` cleanup** — collapses the bespoke `K8sSATrustConfig` into `trustedIssuers` + `allowedClaims` glob map; deletes `k8sSASubPattern` and its single-entry limitations (−252 lines, removal is complete — no orphaned references).
- **JWT signing-key wiring** — `jwt_key.go` loads PEM keys (EC P-256/ES256, RSA≥2048/RS256, PKCS#8), derives `kid` from the RFC 7638 thumbprint, and `createOAuthServer` hard-fails when JWT mode is on without a key, so `enableJWTMode: true` can no longer silently degrade.
- **`dex.allowPrivateIPOIDC`** — SSRF-guard opt-out for OIDC discovery behind internal LBs, mirroring `allowPrivateIPJWKS`.

### Why I'm confident
- **Strong tests** — `jwt_key_test.go` covers EC/RSA/PKCS#8, undersized RSA, wrong curve, ed25519 rejection, missing/invalid/multi-block PEM, and deterministic `kid`. Helm tests cover both `fail` branches plus the success path.
- **Verified all five JWT helm permutations render correctly** — inline key, `existingSecret`, both-set (fails), neither-set (fails), and `existingSecret`-only (wires `jwtSigningKeyFile` to the projected external key). Secret mount path and `jwt-signing-key` key name are consistent across configmap/secret/deployment.
- **Security posture consistent** — both `allowPrivateIP*` knobs default `false`, require explicit opt-in, and emit CWE-918 warnings.

### Non-blocking note
- When `enableJWTMode: true` + `existingSecret` is set, the chart can't verify the external secret actually contains `jwt-signing-key`; a missing key crash-loops the pod at `loadSigningKey` rather than failing at `helm template`. Unavoidable for external secrets and documented in `values.yaml` — just flagging the one late-failure path.

Nice, clean refactor.

PR #788, opened 2026-06-03
fix(netpol): allow egress to Cilium-managed ingress gateway VIP

paurosello:
## Problem

muster fails OIDC discovery against Dex on some clusters with:

```
WARN OIDC discovery failed, retrying in 30s: ... OIDC discovery failed: context deadline exceeded  subsystem=OAuth
```

…while it works fine on others — with the **same chart and the same Cilium config**.

## Root cause (gazelle ✅ vs wallaby ❌)

muster reaches Dex via its public issuer URL, which resolves to the ingress gateway's LoadBalancer VIP. The CNP allowed egress via `toEntities: world` and `toEntities: cluster` on `80/443`. Whether that's sufficient depends on **what kind of LoadBalancer the VIP is**:

| | gazelle (AWS) | wallaby (on-prem) |
|---|---|---|
| `dex.<base>` resolves to | public **AWS ELB** IPs | `10.221.1.14` — **Cilium-managed VIP** |
| Cilium identity of dest | `world` (external) | service VIP → KPR **DNATs to envoy backend `:10443`** |
| Egress rule that matches | `world:443` ✅ | needs backend endpoint `:10443` ❌ (missing) |

With `kube-proxy-replacement=true`, Cilium load-balances the managed VIP **in-datapath**, DNAT'ing `VIP:443` → envoy backend pod `:10443` **before** egress policy is evaluated. So the policy is checked against the backend pod identity on port `10443`, which neither `world:443` nor `cluster:443` matches → SYN dropped → `context deadline exceeded`. On AWS the VIP is an external ELB IP, so traffic egresses as `world` (no DNAT) and `world:443` already covers it.

Verified by reproducing from muster's own network namespace (ephemeral debug container): external `world:443` worked, the gateway VIP `:443` timed out; adding an egress rule to the envoy backend on `10443` fixed it immediately and the warning disappeared on restart.

## Fix

Add a values-driven `networkPolicy.cilium.ingressGateway` egress rule that allows egress to the gateway backend endpoints on their **target** ports, defaulting to the standard Giant Swarm `envoy-gateway` proxies (`10080`/`10443`):

```yaml
networkPolicy:
  cilium:
    ingressGateway:
      endpointSelector:
        k8s:io.kubernetes.pod.namespace: envoy-gateway-system
        app.kubernetes.io/name: envoy
      ports: [10080, 10443]
```

`app.kubernetes.io/managed-by: envoy-gateway` is not present in Cilium security identities on either glean or graveler — Cilium only indexes `app.kubernetes.io/name: envoy` on the proxy pods. The default selector was updated accordingly and verified on glean (Azure/capz).

- **Works on both infra types:** on-prem/managed-VIP clusters get the backend-port rule; cloud-LB clusters keep using the existing unconditional `world` rule (the new rule is a harmless no-op there, and also where the selector matches no pods — non-envoy ingress).
- **Decoupled from `allowClusterIngress`**, which never addressed this (its `cluster:443` rule can't match the DNAT'd `:10443` backend). The misleading comment is corrected.
- Set `ingressGateway: null` to disable.

## Testing

- `helm lint` + `helm template` across `allowClusterIngress` true/false and `ingressGateway` set/null — rule renders/omits correctly.
- Live-validated on wallaby: the equivalent rule made muster reach Dex and cleared the OAuth warning on restart.
- Live-validated on glean (Azure/capz): `app.kubernetes.io/name=envoy` selector fixed OIDC discovery immediately.
- Confirmed gazelle's path (external ELB → `world`) is unchanged.


QuentinBisson (approved):

QuentinBisson:
Verified on glean (Azure/capz, Cilium KPR) and graveler (AWS/capa, Cilium KPR): `app.kubernetes.io/managed-by=envoy-gateway` is **not present** in the Cilium security identity on either cluster — it does not show up in `cilium-dbg endpoint get <pod-ip>` output. The default `endpointSelector` would silently match no pods, leaving the rule as a no-op exactly where it is needed.

The label Cilium does index on the envoy proxy pods is `app.kubernetes.io/name=envoy`. Switching the default to that label fixed OIDC discovery on glean immediately.

Suggested default:

```yaml
ingressGateway:
  endpointSelector:
    k8s:io.kubernetes.pod.namespace: envoy-gateway-system
    app.kubernetes.io/name: envoy
  ports:
    - 10080
    - 10443
```

Tested by applying the equivalent CNP on glean — muster started cleanly with no `OIDC discovery failed` warnings and `/mcp` returned `401 + WWW-Authenticate` immediately.

Note: `app.kubernetes.io/name=envoy` is set by envoy-gateway on its proxy pods and is unlikely to match unrelated pods in `envoy-gateway-system`, so the selector is specific enough in practice.

QuentinBisson (changes requested):
See my previous comment

QuentinBisson (approved):

PR #792, opened 2026-06-03
ci: migrate to push-based release flow

fiunchinho:
## Summary

Replaces the existing release setup with the push-based auto-release flow validated end-to-end in [giantswarm/test-release-please](https://github.com/giantswarm/test-release-please). Mirrors the migration shape from klausctl PR #258.

**muster** had the heaviest old setup: release-please + the colleague's goreleaser-based auto-release + two goreleaser configs (`.goreleaser.yaml` + `.goreleaser.ci.yaml`). All three are removed.

## What changes

**Removed:**
- All workflows from the old release setup (see file list in the diff)
- Any goreleaser configs the colleague's PoC was using

**Added:**
- `.github/workflows/auto-release.yaml` — single workflow that runs on push to `main` (and `release-*` branches for backports). git-cliff computes the next version from conventional commits since the last reachable tag, then `gh release create` atomically creates the tag and the GitHub Release with cliff-rendered notes.
- `cliff.toml` — bump rules, Keep-a-Changelog section mapping (Added/Fixed/Changed/Security), `[remote.github]` integration for PR-link + author rendering, `(?m) \\(#N\\)$` preprocessor to dedupe the squash-merge suffix, `--unreleased --bump` flag combo so the release notes contain only the bumped release.

**Kept:**
- `CHANGELOG.md` — historical reference; future releases publish notes only to GitHub Releases.

## CircleCI

CircleCI is unchanged — muster's pipeline already includes `architect/upload-release-assets` for the muster binary.

## Expected behavior after merge

The merge commit's conventional type (`ci:`) doesn't trigger a bump in git-cliff's default — `ci` maps to "Changed" but only `feat` / `fix` / breaking actually bump. So this PR likely produces no release on its own. The next conventional `fix:` or `feat:` lands as the first release under the new flow.

## Related

- Design + iterative PoC: https://github.com/giantswarm/test-release-please (current state v1.4.3)
- First production migration: https://github.com/giantswarm/klausctl/pull/258


teemow (approved):
Reviewed the push-based migration end-to-end. Clean, well-documented change — the inline rationale (`--unreleased --bump` vs `--latest --bump`, atomic tag+release via `gh release create`, lightweight-tag trade-off, per-branch concurrency, `git describe` reachable-baseline for backports) is excellent.

The earlier `Lint and Test` failure (stale `.goreleaser` references in `Makefile.custom.mk` + the over-length `download_url` line) is resolved in the latest commit — lint is green and `yamllint` passes locally.

One non-blocking nit for a follow-up: in the no-bump / re-run case `git-cliff --unreleased --bump --context` can yield an empty array, and `jq -r '.[0].version'` then emits the literal `null`, which slips past the `[ -z ]` / `[ = last ]` skip guard and would attempt `gh release create null`. `jq -r '.[0].version // empty'` is cheap insurance. Not a blocker if you've confirmed git-cliff returns the current version (not `[]`) on a no-bump push.

fiunchinho:
Thanks for catching this! Applied the `// empty` guard. I couldn't empirically provoke the empty-array case locally (git-cliff returned `[{version: …}]` for every scenario I tried — HEAD at latest tag, HEAD at root commit, chore-only / style-only / non-conventional unreleased commits) — but the fix is cheap insurance and your point about it slipping past both skip-guards is sound.

Also propagated the same fix to:
- giantswarm/klausctl#258
- giantswarm/mcp-kubernetes#450
- giantswarm/klaus-operator#135

So when the migration wave rolls, every consuming repo gets the defensive guard from day one.

PR #818, opened 2026-06-10
chore: drop auto-release.yaml from lint-yaml target

fiunchinho:
## Summary

The `.github/workflows/auto-release.yaml` workflow file has been removed at the platform level, so the `lint-yaml` make target currently fails (`yamllint` errors on the missing file). This drops the reference so `make lint-yaml` works again.

## Test plan

- [ ] `make lint-yaml` succeeds locally
- [ ] `make check` succeeds locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- Drop the removed `.github/workflows/auto-release.yaml` from the `lint-yaml` make target.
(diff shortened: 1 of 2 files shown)


fiunchinho:
Closing —  is still present in this repo, so the lint reference is still valid. Reopen this when the workflow file is actually removed.

fiunchinho:
Correcting previous comment (shell ate the path): closing because `.github/workflows/auto-release.yaml` is still present in this repo, so the lint reference is still valid. Reopen this when the workflow file is actually removed.

PR #820, opened 2026-06-10
fix(helm): render trustedPublicRegistrationRedirectURIs and trustedPublicRegistrationSchemes into configmap

QuentinBisson:
## What

`trustedPublicRegistrationRedirectURIs` and `trustedPublicRegistrationSchemes` were defined in `values.yaml` and validated in `oauth-secret.yaml` (as a gate to satisfy the registration-token requirement), but neither value was written into the `config.yaml` ConfigMap that muster reads at runtime.

External MCP clients using the URI-based DCR gate had their configuration silently dropped on every deploy.

## Fix

Add `{{- with }}` blocks for both fields in `helm/muster/templates/configmap.yaml`, consistent with the pattern used by `trustedAudiences` and `trustedIssuers`.

Add `helm/muster/tests/configmap_test.yaml` covering render-when-set and omit-when-empty for both fields.

### DIFF
--- a/helm/muster/values.schema.json
+++ b/helm/muster/values.schema.json
@@ -503,11 +503,11 @@
             },
             "registrationToken": {
               "type": "string",
-              "description": "Registration token for dynamic client registration"
+              "description": "Shared secret a client must present at the DCR endpoint to register. Supply via existingSecret (key: registration-token) in production."
             },
             "allowPublicClientRegistration": {
               "type": "boolean",
-              "description": "Allow unauthenticated dynamic client registration"
+              "description": "Open registration: any client may register without a token. Rate-limited but still a DoS surface — avoid in production. Default: false."
             },
             "encryptionKey": {
               "type": "boolean",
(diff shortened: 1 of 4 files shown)


fiunchinho (approved):

helm/muster/values.schema.json
```diff
@@ -503,11 +503,11 @@
             },
             "registrationToken": {
               "type": "string",
-              "description": "Registration token for dynamic client registration"
+              "description": "Shared secret a client must present at the DCR endpoint to register. Supply via existingSecret (key: registration-token) in production."
             },
             "allowPublicClientRegistration": {
               "type": "boolean",
-              "description": "Allow unauthenticated dynamic client registration"
+              "description": "Open registration: any client may register without a token. Rate-limited but still a DoS surface — avoid in production. Default: false."
```
QuentinBisson:
```suggestion
              "description": "Open registration: any client may register without a token. Rate-limited but still a DoS surface. Default: false."
```

PR #860, opened 2026-06-18
feat(version): derive version from runtime/debug build info

fiunchinho:
## What

Derive `muster`'s reported version from the Go build info (`runtime/debug`) — which the toolchain stamps from the VCS tag — instead of a hand-maintained literal in `pkg/project/project.go`. Also gitignore the per-arch release binaries.

## Why

When the repo moved off GoReleaser to the push-based auto-release flow, the version stopped being injected. GoReleaser used to set `pkg/project.version` from the tag via `-X` ldflags; architect's `go-build` only injects `gitSHA` and `buildTimestamp`, **never `version`**. To compensate, a version literal was added to `pkg/project` (and was sitting stale at `0.3.13-dev`). The push-based flow does not rewrite source files, so nothing keeps that literal current.

Build info removes the need for the literal entirely:

- **Release builds** run on the tagged commit (the tag push triggers CircleCI, which checks out `CIRCLE_TAG`), so the binaries report the **clean** tag version, e.g. `v0.7.4`.
- **Off-tag builds** report an honest pseudo-version (e.g. `v0.7.5-0.2026...-<sha>`).
- **Tag-less / no-`.git` builds** fall back to the injected `gitSHA`, then `dev`.

`gitSHA` and `buildTimestamp` are still injected by architect's ldflags — unchanged.

## Changes

- `pkg/project/project.go`: `Version()` precedence is now explicit `version` ldflag → build-info `Main.Version` → injected `gitSHA` → `dev`. The literal defaults back to `dev` (left as an `-X` escape hatch). Doc comment corrected to match what architect actually injects.
- `pkg/project/project_test.go`: `buildInfoVersion` is an overridable package var so the fallback chain is unit-tested deterministically.
- `.gitignore`: ignore `/muster-*`. The multi-arch `go-build` loop writes each `muster-<os>-<arch>` into the worktree; without this, every architecture after the first sees an untracked sibling binary and Go marks the build `+dirty`. Verified empirically.

## Safety notes

- architect never passes `-buildvcs=false` and its image ships `git`, so VCS stamping is active.
- A `safe.directory`/dubious-ownership failure would make `go build` **hard-fail** (`error obtaining VCS status`), not silently emit `(devel)` — so a misconfig is loud, never a silently-wrong version. CircleCI's Docker executor runs checkout and build as the same user, so the ownership mismatch shouldn't arise.

## Verification

- `go test ./pkg/project/` passes.
- Local `go build` of this branch reports `v0.7.4+dirty` from build info (`+dirty` only because of the uncommitted edits at build time), confirming the literal is no longer the source.
- **Recommend a single test tag** to confirm the clean version end-to-end in CI before relying on it for the next release.


fiunchinho:
### Validated end-to-end in CI

Proved the approach in `giantswarm/test-release-please` (a throwaway repo) by reproducing architect's `go-build` faithfully on a real tag: shallow detached checkout, `.ldflags` carrying **only** `gitSHA`/`buildTimestamp` (no version), all six arches built into the worktree, then reading the embedded version back with `go version -m`.

**First run failed — and caught a second gotcha this PR now fixes.** The per-arch binaries were correctly ignored, but architect writes two more scratch files into the worktree during `go-build`: `.ldflags` (from `go-test`) and `.platforms`. Untracked, they marked the build `+dirty`, so the embedded version was `v1.6.1+dirty`. Pushed `074227b` to gitignore both.

**Second run passed.** All six architectures embedded the clean tag with no dirty flag:

```
===== app-linux-amd64 =====     mod  …/test-release-please  v1.6.1     vcs.modified=false
===== app-linux-arm64 =====     mod  …/test-release-please  v1.6.1     vcs.modified=false
===== app-darwin-amd64 =====    mod  …/test-release-please  v1.6.1     vcs.modified=false
===== app-darwin-arm64 =====    mod  …/test-release-please  v1.6.1     vcs.modified=false
===== app-windows-amd64.exe === mod  …/test-release-please  v1.6.1     vcs.modified=false
===== app-windows-arm64.exe === mod  …/test-release-please  v1.6.1     vcs.modified=false
PASS: binary embeds clean tag version v1.6.1
```

Confirms: the shallow detached tag checkout resolves a clean semver, the `runtime/debug` fallback works with architect's version-less ldflags, and the gitignore set (`/muster-*`, `.ldflags`, `.platforms`) keeps all six arches clean.

**Caveat:** this ran on a GitHub Actions runner, not the architect CircleCI executor, so it faithfully covers the Go/git mechanics but not the architect-image-specific git-trust (`safe.directory`). Recommend one throwaway test tag on muster itself to close that last gap before a real release depends on it.

QuentinBisson (approved):

teemow (approved):
Approving. Sound change.

- Core mechanism is correct: `debug.ReadBuildInfo().Main.Version` is stamped from the VCS tag as of Go 1.24, and `go.mod` declares `go 1.26.0`, so the assumption holds. CHANGELOG records end-to-end CI validation (clean tag version embedded across all arches with `vcs.modified=false`).
- `Version()` precedence and the `(devel)`/empty guards in `buildInfoVersion` are right; making it an overridable package var gives deterministic unit coverage of the full fallback chain.
- `.gitignore /muster-*` is root-anchored and doesn't shadow anything tracked, so it safely stops the multi-arch go-build loop from marking the build `+dirty`.

Non-blocking: off-tag builds now report a pseudo-version rather than a fixed literal — fine and more honest, just ensure nothing downstream string-matches a clean semver.

PR #911, opened 2026-06-24
chore: temporary build-tree dirtiness diagnostic

teemow:
**Draft / diagnostic only — do not merge.**

Release binaries report `vX.Y.Z+dirty` (e.g. `muster version v0.18.9+dirty`) because Go's build-info VCS stamp records `vcs.modified=true` at link time — the working tree is non-empty when architect's `go-build` links the binaries. The known scratch files (`.ldflags`, `.platforms`, per-arch binaries) are already gitignored and a local repro is clean, so a CI-written artifact is still escaping.

This wires a `vcs-dirty-diagnostic` target as the last `make test` prerequisite (closest consumer-side point to the orb's "Build binaries" step) that prints `git status --porcelain`. The `go-build` job log will show the offending path under the `VCSDIRTY>>` markers.

Once we know the file, revert this and either gitignore it or fix it at the architect-orb layer.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.circleci/workflows.yml
+++ b/.circleci/workflows.yml
@@ -20,7 +20,9 @@ workflows:
         # (-race, integration suites, govulncheck, helm-unittest, CRD-freshness)
         # override their own `make test`. See devctl makefile gen and the
         # make-target-ci-interface ADR.
-        test_target: test
+        # TEMPORARY: test-diag wraps `test` to print git status AFTER go test, to
+        # locate the +dirty culprit. Revert to `test` once identified.
+        test_target: test-diag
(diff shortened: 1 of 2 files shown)


teemow:
Diagnostic complete. Root cause: architect's go-test nancy step writes `./nancy-results.txt` into the repo root between `make test` and the binary link, dirtying the tree so go-build stamps `vcs.modified=true` → `+dirty`. Fixed upstream in giantswarm/architect-orb#857 (write the log to /tmp). No muster change needed.

PR #912, opened 2026-06-24
fix(oauth): connect localMint backends for self-issued OBO sessions

QuentinBisson:
## What

OBO sessions were broken in two layers; both are fixed here.

**Layer 1 — emission (broker/local-mint):** muster-minted OBO tokens had no \`email\` claim. The broker dropped \`SubjectIdentity.Claims\` when projecting into \`MintRequest\`, so \`localMintProvider\` forwarded a bare subject string to \`LocalMintExchanger\` and the minted JWT carried only \`sub\`. mcp-oauth's exchanger already copies \`Claims.Email → AccessTokenClaims.Email\` when the claims are present — the gap was purely on the muster side. Fixed by threading \`SubjectIdentity\` (including \`Claims\`) through \`MintRequest\` and into the \`ExchangerRequest\`.

**Layer 2 — consumption (access-token injector):** a muster self-issued OBO token is emailless at re-presentation — the human identity is in \`sub\`, not an \`email\` claim — and is neither a forwarded nor a trusted-issuer token, so it fell through the emailless branch without firing \`onAuthenticated\`. Without it, the human OBO session never connected its localMint backends, so the agent saw only core tools. Fixed by gating on \`TokenSource == JWT\` in the emailless fall-through: the validated session is bridged into the \`api\` namespace, the subject is set from \`UserInfo.ID\`, and \`onAuthenticated\` fires.

Layer 1 fixes future-minted tokens at the source. Layer 2 handles the re-presentation path and acts as a general fallback for any self-issued token that legitimately has no \`email\` claim.

## Live verification (glean)

After release + auto-pull, a fresh-login UI run should show in the muster audit:
- \`SSO: Connecting … for session ext-<human>\` for the human session, then \`token_issued audience=mcp-kubernetes subject_iss=…muster.glean…\` with an \`act\` claim;
- the agent's tool list includes \`x_kubernetes_*\`, not just \`core_*\`;
- target apiserver audit shows \`impersonatedUser.username=quentin@…\`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- On-behalf-of sessions now connect their localMint backends and advertise their tools. A self-issued OBO bearer (emailless, the human identity in `sub`) reached the dispatch gate after the mcp-oauth v0.18.0 bump but never triggered the per-session backend connect, so the agent saw only core tools. The access-token injector now bridges the validated session into the `api` namespace and fires `onAuthenticated` for self-issued JWTs, so the human session connects its localMint backends and registers their tools.
+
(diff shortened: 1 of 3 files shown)


QuentinBisson:
Closing: the emission-side fix (populate `email`/`groups` claims when minting OBO tokens via localMint) landed directly on main in #883, #886, and #887. That makes the consumption-side workaround here redundant — minted tokens now carry `email`, so the existing code path handles them without the `TokenSourceJWT` special-case.

PR #924, opened 2026-06-25
fix(aggregator): establish localMint backends for human OBO sessions

QuentinBisson:
## What

A human on-behalf-of session presents muster's own self-issued JWT (`sub`=human, `act`=agent), validated as `TokenSourceJWT` with no email claim. Three checks each dropped it, so an OBO session only ever saw the core meta-tools (no backend tools):

1. `createAccessTokenInjectorMiddleware` did not fire `onAuthenticated` for an emailless self-issued JWT, so per-backend SSO bootstrap never ran.
2. `canBootstrapSSO` required an upstream ID token or `TokenSourceOBO` and did not treat an `act`-bearing bearer as a usable subject.
3. `EstablishConnectionWithLocalMint` resolved the local-mint subject from the upstream ID token (which carries no `act`), shadowing the inbound OBO bearer.

This change:
- fires `onAuthenticated` for a self-issued JWT when a session is present;
- accepts a bearer carrying an `act` claim in `canBootstrapSSO`;
- feeds that bearer as the local-mint exchange subject, so the broker re-binds the delegation chain to each backend audience.

Adds `pkg/oauth.HasActClaim`.

## Why

On the glean OBO path, `x_kubernetes_*` / `x_prometheus_*` never appeared for a human agent run; only the 28 core tools did. With these fixes the agent session bootstraps its backends and the per-backend local-mint succeeds. Live on glean the agent lists pods as the human identity.

## Note

Full identity propagation (email and groups in the minted token) also needs giantswarm/mcp-oauth#489; this PR covers the muster-side session and subject-selection fixes and is independent at compile time (builds against mcp-oauth v0.18.5). BDD scenarios for the emailless self-issued-bearer rebind path are a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


QuentinBisson:
Closing: this extended the localMint path, which #947 removes entirely. The forward-token model replaces per-backend issuance: the aggregator forwards the validated inbound bearer byte-identical, and session bootstrap for OBO callers gates on the validated bearer instead of the actor header.

PR #938, opened 2026-07-01
fix(aggregator): refresh oidc-exchange token on the persistent connection

QuentinBisson:
The token-exchange (oidc-exchange) connection built its `Authorization` header from a statically captured exchanged token. mcp-go's continuous listener runs on that connection for the whole session, outliving the exchanged token's fixed lifetime (Dex default 30m) and never traversing the tool-call refresh path (`callToolWithTokenExchangeRetry`). So an idle backend's token expired and the listener looped it (`401 Token validation failed` / `403 invalid session`), leaving the `MCPServer` stuck in `Auth Required`.

Observed fleet-wide on the gazelle central aggregator: nearly every tunnel-backed `<install>-mcp-kubernetes` target sat in `Auth Required` with `lastConnected` days stale, while the only actively-used target stayed `Connected` (its pooled token kept fresh by tool calls). Confirmed not a CA/TLS or audience problem: a freshly exchanged token is a valid RS256 JWT with the trusted `aud` and is accepted by the backend (`userinfo` 200, `kid` present in JWKS). The token was simply expired and never refreshed on the listen connection.

## What changed

- `EstablishConnectionWithTokenExchange` now builds its header via a new `makeTokenExchangeHeaderFunc` that re-exchanges a fresh token once the current one enters the refresh margin before expiry, and keeps the previous token as a fallback on transient failure.
- After `maxConsecutiveTokenFailures` failed re-exchanges (subject token gone / unrefreshable) it evicts the connection so the listener stops instead of looping a dead token, letting the state settle to `Auth Required` cleanly. This mirrors the existing token-forwarding and localMint stale-token handling.
- The re-exchange (subject-token resolution + Dex round-trip) is bounded by a 30s context, well under the 5m refresh margin. Eviction only advances when a re-exchange returns an error, so an unbounded call against a hung Dex would block the header path forever and never settle the connection; the timeout turns a slow endpoint into a normal failure.
- A token whose `exp` cannot be parsed no longer disables refresh for the life of the connection. Both the initial and re-exchanged expiry, when zero, fall back to a bounded refresh interval, so the listener keeps refreshing blind instead of re-stranding on an expired token (the failure mode the refresh exists to prevent).
- No config or backend change: the existing `trustedAudiences` / issuer setup is correct.

## Why re-exchange instead of using a refresh token

Dex's RFC 8693 token-exchange grant returns only an access/ID token plus `expires_in`; it does not issue a refresh token (`ExchangeResult` carries no refresh token), so there is nothing to refresh with. Re-running the exchange from the freshly resolved subject ID token is also the correct behaviour for on-behalf-of: the backend token stays bound to the live user session, so when that session can no longer be refreshed the connection evicts and settles to `Auth Required` rather than a decoupled refresh token keeping a backend session alive for a logged-off user.

## Refresh timing

Refresh is driven by `headerFunc` invocations (tool calls and listener reconnects), not a background timer. On a fully idle backend whose GET listen stream stays open across expiry, recovery is reactive: the first post-expiry reconnect re-exchanges instead of looping the dead token. The margin-based path refreshes ahead of expiry whenever a header call lands in the window.


internal/aggregator/connection_helper.go
```diff
@@ -788,18 +789,64 @@ func EstablishConnectionWithTokenExchange(
...
+		// failure counter, so a slow endpoint must surface as a normal failure.
+		ctx, cancel := context.WithTimeout(context.Background(), tokenReexchangeTimeout)
+		defer cancel()
+
+		freshID := getIDTokenForForwarding(ctx, sessionID, musterIssuer, a.sessionRefresher())
+		if freshID == "" {
+			return "", time.Time{}, fmt.Errorf("no subject ID token available for re-exchange")
+		}
+		if expired, expErr := pkgoauth.IsExpired(freshID); expired {
+			return "", time.Time{}, fmt.Errorf("subject ID token expired: %w", expErr)
```
teemow: Nit: malformed error string on the genuine-expiry branch. `pkgoauth.IsExpired` returns `(true, nil)` when the token is *actually* past expiry (it only returns `(true, err)` when unparseable / no `exp`). So on the normal case `expErr` is nil and `%w` renders `subject ID token expired: %!w(<nil>)`. Wrap only when `expErr != nil`, or use a plain message here.

internal/aggregator/connection_helper.go
```diff
@@ -905,17 +952,111 @@ func emitTokenExchangeEvent(serverName, namespace string, success bool, errorMsg
...
+// block indefinitely. A zero expiry (token carries no parseable exp) is treated
+// as tokenExchangeFallbackRefreshInterval from now, so refresh keeps firing
+// blind rather than latching off and re-stranding the listener.
+//
+// mcp-go invokes the returned closure concurrently on a single connection: the
+// continuous-listening GET runs in its own goroutine (listenForever) while tool
+// calls invoke it from their own goroutines, and both reach headerFunc via
+// sendHTTP. The mutex is therefore required for correctness, not just defensive;
+// it also serialises the bounded reexchange so a single refresh serves any
+// concurrent caller (single-flight) rather than each firing its own Dex round-trip.
```
teemow, line 983: This corrected understanding (headerFunc is invoked concurrently, mutex required) directly contradicts the doc comments on the sibling `makeTokenForwardingHeaderFunc` and `makeLocalMintHeaderFunc`, which still claim "called sequentially per connection, so no lock needed" and mutate `consecutiveFailures` / `staleEvicted` / `lastWarnTime` / `hadToken` without a mutex. Under a concurrent listener + tool call those counters race (benign-ish skew / possible double-fire of `onStaleToken`, but `-race` would flag it). Pre-existing, but this PR is what proves the assumption wrong — worth a follow-up to add a mutex to the siblings.

teemow (commented):
Reviewed the branch in a throwaway worktree: all 6 new tests pass, `go vet` clean, `-race` clean. Root cause, the 30s re-exchange bound, and the concurrency correction (commit 4) all check out — I traced mcp-go v0.55.1 and confirmed `listenForever` (own goroutine) and tool calls both reach `headerFunc` via `sendHTTP` concurrently, so the mutex is genuinely required. Three notes inline; none block the fix.

Extra follow-up that has no line to anchor to: the self-refresh only protects the initial connection. The tool-call / background-refresh path (`exchangeTokenAndCreateClient` in `server.go`, ~L2862) still builds a **static** header func. Recurrence path: a backend gets one tool call while expiring-soon → `triggerBackgroundTokenRefresh` swaps in a static-header client → backend goes idle → the new listener re-strands exactly as before. The common purely-idle case is fully fixed; consider reusing `makeTokenExchangeHeaderFunc` there too, or documenting it as a known limitation.

internal/aggregator/connection_helper.go
QuentinBisson: Agreed, and confirmed against mcp-go v0.55.1: `listenForever`→`sendHTTP`→`headerFunc` and `SendRequest`→`sendHTTP`→`headerFunc` both hit the closure concurrently with no lock inside mcp-go, so the two sibling funcs genuinely race. Kept it out of this PR (pre-existing, other auth modes, wants its own concurrent `-race` regression test) and tracked it in #939.

QuentinBisson:
Follow-up: #943 applies the same mutex fix to the sibling `makeTokenForwardingHeaderFunc` and `makeLocalMintHeaderFunc` closures (#939), which this PR's concurrency finding surfaced. Independent of this PR (separate funcs, no file conflict).

PR #940, opened 2026-07-01
fix(mcpserver): ignore runtime token-exchange creds in ConfigurationChanged

fiunchinho:
## What

MCPServers using RFC 8693 token exchange restart on **every reconcile pass** (~every 10-15s), even though their CR spec never changes (`generation` stays 1). Each restart drops the server to pending-auth, so a tool call landing in the restart/reauth window fails with `authentication required` / `user not authenticated to server`.

Fixes the churn tracked in giantswarm/giantswarm#37060 (observed live on the `wallaby` management cluster).

## Root cause

`Service.ConfigurationChanged` (`internal/services/mcpserver/service.go`) compared the two auth configs with `reflect.DeepEqual(cur.Auth, newDef.Auth)`.

`TokenExchangeConfig.ClientID` and `ClientSecret` are tagged `json:"-" yaml:"-"` and are **resolved from a Kubernetes Secret at runtime**, then stored on the running service's definition. A definition rebuilt from the CR has them empty. So once credentials load, `cur.Auth` (populated) and `newDef.Auth` (empty creds) always differ → `ConfigurationChanged` returns `true` on every pass → the reconciler updates + restarts the service in a loop.

## Change

- New helper `authConfigChanged` compares auth via a normalized copy (`normalizeAuthForComparison`) that clears the runtime-only `ClientID`/`ClientSecret` on a **copy** of the token-exchange config before `reflect.DeepEqual`. The inputs are never mutated, and every spec-derived field (including any added later) still participates in the comparison — only the two resolved-credential fields are excluded.
- Two new `TestConfigurationChanged` table cases:
  - `auth token-exchange runtime credentials differ only` → **no change** (the regression).
  - `auth token-exchange spec field changed` (Scopes) with creds also differing → **change still detected**.

## Testing

- `go build ./...` — OK
- `go test ./internal/services/mcpserver/` — OK (all cases pass, including the two new ones)
- `go vet ./internal/services/mcpserver/` — clean
- `gofmt` — clean

Refs giantswarm/giantswarm#37060


fiunchinho:
### Code review

Found 1 issue:

1. The fix ignores the runtime-resolved `ClientID`/`ClientSecret` but misses a third field that is also mutated at runtime on the *same* `Auth` pointer: `TokenExchange.Scopes`. When a server sets `requiredAudiences`, the aggregator appends the audience scopes in place onto the shared `TokenExchangeConfig` during connection:

https://github.com/giantswarm/muster/blob/6ac5e19a65708e627327fc770b8fc9dcdaff72ae/internal/aggregator/connection_helper.go#L725-L739

That struct is the very object `ConfigurationChanged` compares: the running definition's `Auth` pointer is handed to the aggregator unchanged (`orchestrator.go` → `AuthConfig: definition.Auth`, stored without copying in `registry.go`), which is exactly why the `ClientID`/`ClientSecret` churn this PR fixes was observable in the first place. Since `normalizeAuthForComparison` clears only the two credential fields and leaves `Scopes` intact, once a connection appends audiences the running definition's `Scopes` (e.g. `openid profile email groups audience:...`) permanently differs from the CR-derived `Scopes` (`openid profile email groups`), so `authConfigChanged` keeps returning `true` and the same restart-every-~10-15s churn persists for any MCPServer that combines `tokenExchange` with `requiredAudiences` — plausibly the reported scenario, since audience scopes exist for remote-cluster OIDC.

https://github.com/giantswarm/muster/blob/6ac5e19a65708e627327fc770b8fc9dcdaff72ae/internal/services/mcpserver/service.go#L438-L448

The robust fix is to stop mutating the shared registry pointer in `connection_helper.go` (copy `TokenExchangeConfig` before appending scopes, as `server.go` already does elsewhere); alternatively, also normalize `Scopes` here. Worth adding a regression test for the `requiredAudiences` case, since neither new test exercises it.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

fiunchinho:
Agreed — verified the shared-pointer path and you're right, the credential-only normalization was incomplete.

`orchestrator.go` assigns `AuthConfig: definition.Auth` (the registry definition's `*MCPServerAuth`) straight into `serverInfo`, and `TokenExchange` is itself a `*TokenExchangeConfig`, so `connection_helper.go` was mutating the exact object `ConfigurationChanged` compares — not just for `ClientID`/`ClientSecret` but for the appended audience `Scopes` too. Since every affected install sets `requiredAudiences: [dex-k8s-authenticator]`, the `Scopes` divergence alone kept the ~10-15s churn going despite the first fix.

Fixed at the root (`f17d001`): the connection path now applies the per-connection mutations (credentials + audience scopes) to a **value copy** of the config, mirroring the existing pattern in `server.go`, so the shared registry definition is never modified. I went with the copy rather than also normalizing `Scopes` in the comparison, since `scopes` is a real CR field and a genuine change to it *should* still trigger a restart. Kept the `ClientID`/`ClientSecret` normalization as defense-in-depth (they're `json:"-"` runtime-only fields). Added regression tests covering the `requiredAudiences` case and asserting the shared config is left untouched.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ All notable changes to this project will be documented in this file.
...
....go` mutated `serverInfo.AuthConfig.TokenExchange` in place — but `AuthConfig`/`TokenExchange` are pointers shared with the registry's MCPServer definition (`orchestrator.go`: `AuthConfig: definition.Auth`), which is the exact object `MCPServerReconciler.ConfigurationChanged` compares against the CR. Each connection wrote the runtime-resolved `ClientID`/`ClientSecret` (tagged `json:"-" yaml:"-"`) and, for any server with `requiredAudiences`, appended cross-client audience scopes onto the shared `Scopes`. The stored definition then permanently differed from the CR-derived one, so the reconciler tore down and reconnected the MCPServer every ~10-15s; a tool call landing in the restart/reauth window failed with `authentication required` / `user not authenticated to server`. The connection now applies these per-connection mutations to a value copy of the config (mirroring `server.go`), so the shared definition is never modified. As defense-in-depth, `ConfigurationChanged` also ignores the runtime-only `ClientID`/`ClientSecret` credential fields while still detecting every spec change (including a genuine `scopes` change). ([#37060](https://github.com/giantswarm/giantswarm/issues/37060))
```
QuentinBisson: The entry carries the full root-cause narrative (pointer sharing, `orchestrator.go` internals, the churn mechanism). CHANGELOG should say what the consumer gets; the why is already well captured in the PR body. Suggest trimming:

```suggestion
- MCPServers using RFC 8693 token exchange no longer restart on every reconcile pass (~every 10-15s), which previously caused intermittent `authentication required` / `user not authenticated to server` errors on tool calls landing in the restart window. ([#37060](https://github.com/giantswarm/giantswarm/issues/37060))
```

internal/aggregator/connection_helper.go
```diff
@@ -699,52 +729,49 @@ func EstablishConnectionWithTokenExchange(
...
-		}
+	// Build a value copy of the TokenExchange config with the per-connection
+	// mutations (resolved credentials + required-audience scopes) applied. We must
+	// NOT mutate serverInfo.AuthConfig.TokenExchange in place: AuthConfig and
+	// TokenExchange are pointers shared with the registry definition (see
+	// orchestrator.go: `AuthConfig: definition.Auth`), which is the exact object
+	// MCPServerReconciler.ConfigurationChanged compares. In-place mutation of the
+	// runtime-only ClientID/ClientSecret and the appended audience Scopes makes the
+	// stored definition permanently differ from the CR, causing a restart on every
+	// reconcile. This mirrors the value-copy pattern already used in server.go.
```
QuentinBisson: This block duplicates the rationale already on `buildConnectionTokenExchangeConfig`'s doc comment. Keep the invariant pointer at the call site, drop the repeated bug/ticket narrative:

```suggestion
	// Apply per-connection mutations to a value copy; never mutate
	// serverInfo.AuthConfig.TokenExchange in place, it is shared with the registry
	// definition (see buildConnectionTokenExchangeConfig).
```

QuentinBisson (approved):
Correct and minimal fix. Root cause (in-place mutation of the shared registry definition pointer) is addressed at the connection site with a value copy, and the `ConfigurationChanged` credential masking is sound defense-in-depth. Verified nothing downstream reads the previously-mutated fields. The two inline suggestions are non-blocking, and the DRY consolidation is tracked in #942.

PR #947, opened 2026-07-02
feat(oauth): forward the validated token to backends, remove localMint and X-Actor-Token

QuentinBisson:
Phase 2 of the OBO single-token refactor (muster side). The aggregator stops re-issuing a per-backend token and instead forwards the caller's validated token unchanged, so a backend validates the forwarded token directly and reads its nested `act` delegation chain. On-behalf-of tokens carry `aud` = muster's resource identifier; a directly presented trusted-issuer token carries that issuer's audience, and both flow through byte-identical.

## What changed

**Forward-token data plane.** Both the connect-time client builder (`newTokenForwardingClient`, shared by discovery and the pool-miss tool-call path) and the per-request header func resolve tokens in the same order: (1) the validated inbound bearer from the request context, forwarded byte-identical (agent OBO sessions and JWT-mode human sessions; opaque bearers are never forwarded), (2) the session's stored upstream ID token (listen stream, opaque-token sessions), (3) the token captured at connect. A resolution counts as failed only when it bottoms out on an expired fallback and a last-resort in-place session refresh recovers nothing, so sessions with no token-store entry (agent OBO callers) never warn-loop toward eviction while their token is valid, and opaque-token sessions whose stored ID token expired recover without an eviction cycle. The failure counter is atomic, so requests that carry their own bearer never wait behind a store lookup or an in-flight refresh; sessions with no upstream refresh chain keep the stale-token eviction from #938/#943.

**Session bootstrap and auth gate.** The bootstrap gates on `IDToken || forwardable bearer`; an opaque bearer does not count, so the broken-refresh-chain eviction in `onAuthenticated` still fires for opaque-token sessions. `onAuthenticated` also fires for emailless authenticated sessions, which is what lets an agent session establish its backends. Token-forwarding servers accept a forwardable validated bearer in place of a per-server auth-store entry at the tool-call gate, so a stale-token eviction that revoked the entry cannot fail a returning caller that presents a fresh token.

**localMint removed end to end.** The `auth.localMint` MCPServer CRD field and its CEL rules (CRDs regenerated), the `local-mint` broker target type and target `.type` key, `LocalMintExchanger` wiring, the `BackendTokenMinter` seam (`internal/api/token_minter.go`, `internal/server/api_adapter.go`), `EstablishConnectionWithLocalMint` / `makeLocalMintHeaderFunc` / `ShouldUseLocalMint` and every branch site.

**X-Actor-Token removed.** The actor token is presented once as the RFC 8693 `actor_token` parameter at `/oauth/token`; `/mcp` requests carry only the issued OBO token. Header constant, extraction, context plumbing, `CallerTokens.Actor`, and the test-harness senders are gone.

**Broker collapsed to one behavior.** With `local-mint` gone, the `CredentialProvider` registry/factory indirection had one implementation left; `BrokerExchanger.Exchange` now performs the downstream Dex exchange directly (credentials stamped via `api.TokenExchangeConfig.WithResolvedRuntime`). `BrokerTargetType` and `delegateToSelf` are removed (mcp-oauth defaults `aud` to the resource identifier natively). Broker targets are validated at startup: a target without `dexTokenEndpoint` fails startup naming the audience, and downstream exchange errors are wrapped with the audience.

**JWKS trust pool.** mcp-oauth's JWKS clients no longer read the CA installed on `http.DefaultTransport`; the pool must be passed explicitly. The process trust pool that `--extra-ca-file` augments is now handed to `Config.JWKSRootCAs` and every `TrustedIssuer.RootCAs`, so trusted-issuer subject validation against an internal-CA IdP verifies instead of failing with `certificate signed by unknown authority`. TLS-only: issuer authorization stays the `trustedIssuers` allowlist.

**BDD.** The six `localmint-*` scenarios are replaced by `obo-token-forward` (real protocol traffic: subject+actor exchange at `/oauth/token`, `/mcp` with the OBO bearer, backend validates against muster's JWKS and echoes `sub`/`act` from the token it received) and `obo-token-forward-deny` (untrusted issuer refused at both `/oauth/token` and `/mcp`). Full suite: 178/178 green.

## Dependency

mcp-oauth is pinned to the tagged `v1.0.1`: `WorkloadExchangeSubjectToken`, `LocalMintExchanger`, `EnableWorkloadTokenExchange`, and `DelegationDefaultResource` are gone and the self-issued exchange path replaces them. v1.0.1 also carries the permissive-JWKS cross-host redirect guard, the unverified-email issuance gate (the exchange refuses a subject asserting an email without `email_verified: true`, restoring the fail-closed check the removed localMint path enforced), and the self-issued exchange hardening (actor re-attachment gate, DPoP subject binding, resource allowlist). The whole stack (muster, mcp-kubernetes, mcp-prometheus) shares this one revision.

## Deployment note

JWT-mode installs: a human session's inbound bearer is a muster-signed JWT, so every `forwardToken` backend switches from receiving the upstream Dex ID token to receiving muster's token. Backends that validate the Dex issuer must be reconfigured to trust muster's issuer/JWKS before (or together with) this muster release, or their human path breaks. Opaque-token installs are unaffected.

Validated end to end on glean: the OBO exchange, forward, and backend-validation path is green there, and the deny path (untrusted issuer refused at both boundaries) is confirmed.

Part of a stacked rollout (mcp-oauth → muster → backends → agentic-platform → configs), merged bottom-up. Closes #924's line of work: it extended the removed localMint path.

## MC impersonation removal (glean)

The interim audience-injection middleware (`withAgentExchangeAudience`, commit dffd0b8) was reverted on this branch before shipping. Removing apiserver impersonation on the MC is instead handled outside this PR by the existing per-backend `tokenExchange` machinery: the glean mcp-kubernetes MCPServer switches to `auth.mode: exchange` so muster exchanges the session token at glean Dex for a Dex-signed `aud=dex-k8s-authenticator` token (giantswarm-configs#750 + shared-configs#688). No aggregator changes were needed; kagent-dev/kagent#2106/#2107 are no longer prerequisites.


QuentinBisson:
@teemow I'm waiting for your agent to review this one :)

teemow (changes requested):
The removals in this PR are all welcome and should land: localMint end to end, the X-Actor-Token header contract, the CredentialProvider indirection, plus the JWKS CA-pool fix and broker-target startup validation. But the replacement trust model is wrong and can't ship. Two blocking changes:

## 1. forwardToken must keep sending the upstream dex id_token — unchanged

Muster is an aggregator, not an identity provider. It has no authority to assert SSO identity to downstream MCP servers; that is dex's job. Downstream servers validate tokens against their own IdP (dex), and forwardToken with OBO has to work exactly like forwardToken without OBO: the upstream dex id_token goes downstream, byte-identical, in every mode.

This PR instead resolves the validated inbound bearer first, so on JWT-mode installs every forwardToken backend silently switches from receiving the dex id_token to receiving a muster-signed JWT — the deployment note even asks backends to be "reconfigured to trust muster's issuer/JWKS". No backend should ever be configured to trust muster as an issuer.

Concretely, restore id_token-only resolution (main's behavior):
- drop the `forwardableBearer` fast path in `newTokenForwardingClient` and `makeTokenForwardingHeaderFunc` (internal/aggregator/connection_helper.go)
- drop the tool-call auth-gate bypass for forwardable bearers in `getOrCreateClientForToolCall` (internal/aggregator/server.go)
- drop the `canBootstrapSSO` change accepting a forwardable bearer (internal/aggregator/sso_session.go) and the emailless `fireOnAuthenticated` bootstrap (internal/server/oauth_http.go)

## 2. Remove JWT mode instead of building on it

`enableJWTMode` / `jwtSigningKeyFile` / the self-issued exchange make muster sign identity-bearing access tokens. That is the same wrong authority again and it has to go, not be extended. Sessions whose only credential is a muster-signed OBO token cannot use forwardToken backends by design — an agent acting on behalf of a user needs a dex-issued token, obtained by exchanging at dex (exactly what glean mcp-kubernetes now does with `auth.mode: exchange`; that pattern is the answer, not muster-issued JWTs).

Note the brokered exchange path (confidential client + audience allowlist → downstream dex token) does not depend on JWT mode and stays — it issues dex tokens, which is the correct authority.

This also closes a concrete hole in the current state: muster never sets mcp-oauth's `TokenExchangeAllowedResources`, so the credential-less self-issued exchange will sign a muster JWT for **any** requested `resource`/aud for any caller holding a trusted-issuer token. Removing the self-issued path removes that surface entirely.

## Suggested sequencing

This PR lands the removals with id_token-only forwarding restored (mostly reverting the forwarding/bootstrap commits on this branch). JWT-mode removal (config, helm values/schema, signing-key wiring, BDD harness `muster_broker` JWT bits, mcp-oauth self-issued surface) follows immediately after — happy to help scope that.

teemow:
Review found one hardening gap, fixed in 200c763b before merge: mcp-oauth v1.0.1's `TokenExchangeAllowedResources` allowlist was never set by muster, and the empty default disables the check — so any caller holding a valid trusted-issuer token could use the credential-less self-issued exchange at `/oauth/token` to obtain a muster-signed JWT with an arbitrary `aud` (RFC 8707 `resource`). The allowlist is now pinned to muster's `resourceIdentifier`, so the self-issued path mints only muster-audience tokens; other audiences must go through the brokered path (client auth + per-client audience allowlist). Pinned by a unit test and a new `foreign-resource-denied` step in `obo-token-forward-deny`. Full BDD suite 178/178 green with the change.

teemow:
Correction on the process and the direction: this PR was admin-merged past Timo's request-changes review, which was a mistake. The ruling in that review stands: **dex is the sole SSO authority; muster is not an IdP and must never sign tokens; no backend may trust a muster issuer/JWKS; the tokenExchangeBroker is for Backstage access only.**

#968 (merged, released as **v1.0.0**) implements it: JWT mode removed end to end, the self-issued RFC 8693 exchange permanently disabled, `forwardToken` forwards only IdP(dex)-issued tokens, and the `obo-token-forward*` scenarios replaced by `dex-token-forward*` (dex mints the act chain; the backend validates against dex's JWKS). This PR's localMint/X-Actor-Token removal and forwarding-refactor survive; its muster-signed-token direction and the "backends must trust muster's issuer/JWKS" deployment note are reverted/superseded. giantswarm-configs#750 and shared-configs#688 carry do-not-merge comments pending rework against the dex-only model.

PR #950, opened 2026-07-03
docs(crd): add example CRs for the docs CRD reference

marians:
## What

Adds correctly-named, single-document example CRs under `docs/cr/`:

- `docs/cr/muster.giantswarm.io_v1alpha1_workflow.yaml`
- `docs/cr/muster.giantswarm.io_v1alpha1_mcpserver.yaml`

## Why

We're publishing muster's CRDs on the [Giant Swarm CRD reference](https://docs.giantswarm.io/reference/platform-api/crd/)
(companion PR in `giantswarm/docs`). The generator
([`crd-docs-generator`](https://github.com/giantswarm/crd-docs-generator))
only attaches an example to a CRD page when it finds a file named **exactly**
`{group}_{version}_{singular}.yaml` sitting **directly** in a configured
`cr_paths` directory, with **one resource per file**.

Our existing `examples/` files don't match that convention (wrong names, and
`mcpserver-example.yaml` bundles six resources in one file), so these dedicated
files are added for the docs generator. Content is adapted from the existing
`examples/` so it stays representative.

## Notes

- Once merged, this needs a release so `giantswarm/docs` can pin its
  `commit_reference` to a tag that contains `docs/cr/`.
- `WorkflowExecution` is intentionally not included — it's a controller-created
  runtime record, not a user-authored resource. Easy to add later if we want it
  on the reference too.

### DIFF
--- a/docs/cr/muster.giantswarm.io_v1alpha1_workflow.yaml
+++ b/docs/cr/muster.giantswarm.io_v1alpha1_workflow.yaml
@@ -24,17 +24,17 @@ spec:
       args:
         name: "{{.input.app_name}}"
         tag: "{{.input.environment}}-latest"
-      store: true
+      output: true
     - id: deploy_service
       tool: kubectl_apply
       args:
         name: "{{.input.app_name}}-{{.input.environment}}"
         image: "{{.results.build_image.image_id}}"
         replicas: "{{.input.replicas}}"
-      store: true
+      output: true
     - id: verify_deployment
       tool: health_check
       args:
         service_name: "{{.results.deploy_service.name}}"
         timeout: "5m"
-      store: true
+      output: true
--- a/docs/cr/muster.giantswarm.io_v1alpha1_workflow.yaml
+++ b/docs/cr/muster.giantswarm.io_v1alpha1_workflow.yaml
@@ -1,0 +1,40 @@
+apiVersion: muster.giantswarm.io/v1alpha1
+kind: Workflow
+metadata:
+  name: deploy-application
+  namespace: default
+spec:
+  description: "Deploy application to production environment with validation"
+  args:
+    app_name:
+      type: string
+      required: true
+      description: "Name of the application to deploy"
+    environment:
+      type: string
+      default: "production"
+      description: "Target deployment environment"
+    replicas:
+      type: integer
+      default: 3
+      description: "Number of application replicas"
+  steps:
+    - id: build_image
+      tool: docker_build
+      args:
+        name: "{{.input.app_name}}"
+        tag: "{{.input.environment}}-latest"
+      store: true
+    - id: deploy_service
+      tool: kubectl_apply
+      args:
+        name: "{{.input.app_name}}-{{.input.environment}}"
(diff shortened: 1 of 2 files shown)


docs/cr/muster.giantswarm.io_v1alpha1_workflow.yaml
```diff
@@ -0,0 +1,40 @@
...
+      type: integer
+      default: 3
+      description: "Number of application replicas"
+  steps:
+    - id: build_image
+      tool: docker_build
+      args:
+        name: "{{.input.app_name}}"
+        tag: "{{.input.environment}}-latest"
+      store: true
```
teemow: `store` is deprecated in favor of `output` (see the field docs in the CRD schema). Since this file is what the public CRD reference will show, better to use the current field:

```suggestion
      output: true
```

docs/cr/muster.giantswarm.io_v1alpha1_workflow.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - id: deploy_service
+      tool: kubectl_apply
+      args:
+        name: "{{.input.app_name}}-{{.input.environment}}"
+        image: "{{.results.build_image.image_id}}"
+        replicas: "{{.input.replicas}}"
+      store: true
```
teemow:
```suggestion
      output: true
```

teemow (commented):
Checked both files against the CRD schemas in `helm/muster/crds/` and against the crd-docs-generator source — the filename convention (`{group}_{version}_{singular}.yaml`, one resource per file, directly in a `cr_paths` dir) matches exactly, and the MCPServer example passes all schema + CEL admission rules for a stdio server.

One thing worth fixing before this lands on the public reference: the Workflow example uses `store: true`, which the CRD itself marks as a deprecated alias for `output` ("Prefer Output"). Since these examples exist specifically to be shown on docs.giantswarm.io, they'd teach every reader the deprecated field. Inline suggestions attached — the semantics are identical (`output` controls whether the step result appears in the returned document; results are always referenceable by later steps either way).

The stale `store` usage in `examples/workflow-example.yaml` (which this was copied from) could be cleaned up in a follow-up, no need to grow this PR.

docs/cr/muster.giantswarm.io_v1alpha1_workflow.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - id: verify_deployment
+      tool: health_check
+      args:
+        service_name: "{{.results.deploy_service.name}}"
+        timeout: "5m"
+      store: true
```
teemow:
```suggestion
      output: true
```

marians:
Thanks for the thorough review! Applied all three suggestions — the Workflow example now uses `output: true` instead of the deprecated `store`.

Good call on keeping it out of the deprecated field on the public reference. I'll clean up the stale `store` usage in `examples/workflow-example.yaml` in a separate follow-up PR as you suggested, so this one stays focused.

teemow (approved):

Issue #965, opened 2026-07-09
Remove /oauth/token audience-injection workaround once kagent #2106/#2107 merge


QuentinBisson:
## What

muster injects the `dex-k8s-authenticator` broker audience for **agent token-exchange** requests that arrive with no `audience`, because kagent's STS plugin passes `nil` (`kagent-dev/kagent` `go/adk/pkg/sts/plugin.go:183`). This lets muster mint the **Dex-signed** apiserver token (`iss=dex.<mc>`, `aud=dex-k8s-authenticator`) via its broker cross-client exchange, so mcp-kubernetes can pass it through natively (`--downstream-oauth`) and we can **drop MC impersonation now** instead of waiting for kagent.

## Why it's temporary

The clean way to scope the exchanged token is `KAGENT_STS_AUDIENCE`, added upstream by:
- kagent-dev/kagent#2106 (Go ADK)
- kagent-dev/kagent#2107 (Python ADK)

Both are open with no merge ETA. Until they land, kagent cannot scope the token, so muster injects the audience server-side as a bridge.

## When #2106/#2107 merge, do

1. Set `KAGENT_STS_AUDIENCE` (to the muster broker target that mints `dex-k8s-authenticator`) in the agent config.
2. Delete the muster `/oauth/token` audience-injection middleware + its `TEMPORARY WORKAROUND` comment (`internal/server/oauth_http.go`).
3. Confirm the minted token is unchanged: `iss=dex.<mc>`, `aud=dex-k8s-authenticator`, `sub=<email>`, `groups` present; apiserver authenticates it with no `Impersonate-*` headers.

## Reference

gazelle already runs the passthrough model live for the human path (`gazelle-mcp-kubernetes`: `forwardToken:true` + `requiredAudiences:[dex-k8s-authenticator]`, mcp-kubernetes `--downstream-oauth`).

QuentinBisson:
Workaround implemented in #966 (`internal/server/oauth_http.go` `withAgentExchangeAudience` + `TokenExchangeBrokerConfig.DefaultAgentAudience`). This issue tracks its removal once kagent-dev/kagent#2106 + #2107 merge.

QuentinBisson:
[...]

PR #966, opened 2026-07-09
feat(oauth): inject agent exchange audience for apiserver-valid backend tokens

QuentinBisson:
Temporary workaround so we can remove MC impersonation now, without waiting for kagent.

## Problem

The glean apiserver trusts Dex only (`aud=dex-k8s-authenticator`). kagent's STS exchange presents subject+actor but passes a **nil audience** (`kagent-dev/kagent` `sts/plugin.go:183`), so muster self-issues `aud=<resourceIdentifier>`, which the apiserver can't validate, forcing mcp-kubernetes to impersonate.

## Change

For an on-behalf-of token-exchange request (grant_type=token-exchange with an `actor_token`) that arrives with **no `audience`**, muster injects `TokenExchangeBroker.DefaultAgentAudience`. That routes the exchange to the matching broker target, so muster mints a **Dex-signed `dex-k8s-authenticator`** token. mcp-kubernetes passes it through natively (`--downstream-oauth`), so impersonation on the MC can be removed. `act` is still logged at the broker exchange.

The injection only fires when an `actor_token` is present and no explicit `audience` was requested, so a caller that scopes its own token is never overridden. Disabled when `DefaultAgentAudience` is empty.

## Temporary — tracked by #965

kagent-dev/kagent#2106 (Go) + #2107 (Python) add `KAGENT_STS_AUDIENCE`. Once merged, kagent scopes the token itself; set that env var and delete this middleware + the `DefaultAgentAudience` field.

Depends on the forward-token stack (#947); based on `obo/forward-token`.

gazelle already runs this passthrough model live for the human path.

### DIFF
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -373,6 +373,17 @@ type TokenExchangeBrokerConfig struct {
 	// secret refs that do not set an explicit namespace. Populated from the
 	// muster namespace by the serve command; not user-facing config.
 	DefaultSecretNamespace string `yaml:"-"`
+
+	// DefaultAgentAudience is a TEMPORARY WORKAROUND (giantswarm/muster#965).
+	// When an agent on-behalf-of exchange (grant_type=token-exchange with an
+	// actor_token) arrives with no `audience`, muster injects this audience so
+	// the exchange routes to the matching Targets entry and mints a Dex-signed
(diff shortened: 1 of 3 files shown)


QuentinBisson:
Auto-closed as merged because the commit was fast-forwarded onto obo/forward-token — it now rides in #947. No separate merge to main.

PR #997, opened 2026-07-28
feat(helm): serviceMonitor.enabled implies the prometheus metrics exporter

fiunchinho:
## Why

Muster on the gazelle MC was silently exporting no metrics: the umbrella chart enabled the ServiceMonitor toggle but not the `prometheus` exporter, and that half-configuration is a silent no-op — no `/metrics` endpoint, no metrics port, no ServiceMonitor rendered.

The two knobs are redundant for the dominant use case: asking for a ServiceMonitor unambiguously means "serve /metrics".

## What

- New `muster.effectiveMetricsExporter` helper: normalizes `muster.observability.metrics.exporter` and appends `prometheus` when `muster.observability.metrics.prometheus.serviceMonitor.enabled` is true.
- `OTEL_METRICS_EXPORTER` and the resource-attributes gate in the deployment now use the effective list, so the implication reaches the process, not just the k8s objects.
- `muster.prometheusExporterEnabled` (gates the metrics container/Service ports, prometheus env, and ServiceMonitor) is now derived from the effective list.

With this, a single value enables the whole scrape path:

```yaml
muster:
  observability:
    metrics:
      prometheus:
        serviceMonitor:
          enabled: true
```

Explicit exporter lists are preserved: `exporter: "otlp"` + the toggle yields `otlp,prometheus`. Exact-element matching is unchanged (`fakeprometheus` still doesn't count).

## Testing

- `helm unittest`: servicemonitor + deployment suites updated/extended for the implication (29 passing). The pre-existing `rbac_test.yaml` failure on `main` (missing `workflowexecutions`) is unrelated and untouched.
- `make helm-lint` passes.
- Smoke: `helm template --set muster.observability.metrics.prometheus.serviceMonitor.enabled=true` renders the ServiceMonitor, metrics Service/container ports, and `OTEL_METRICS_EXPORTER=prometheus`.


marians (approved):
LGTM

Issue #951, opened 2026-07-04
Complete mcp-oauth v1 migration (localmint OBO scenarios fail: client_id now required in token exchange)


teemow:
## Problem

Renovate PR #949 bumped `github.com/giantswarm/mcp-oauth` to v1. The build compiled, but 6 localmint scenarios failed in `make test`:

- localmint-broker-negative-audience, localmint-broker-obo-claims, localmint-broker-downstream-reject, localmint-broker-obo-downstream (failed)
- localmint-obo-discovery, localmint-backend-tools-visible (error)

## Resolution (2026-07-05)

Only **one** real v1 behavior change was involved: CA trust is no longer read from `http.DefaultTransport`. mcp-oauth v1's permissive JWKS / token-exchange / OIDC-discovery clients need an explicit `*x509.CertPool`. The fix threads muster's `--extra-ca-file` pool through `config.OAuthServerConfig` (new `pkg/tlsutil.LoadCAPool`, single pool shared by the DefaultTransport install and the explicit clients).

The suspected second change — "token exchange now rejects requests without `client_id`" — was a **misdiagnosis**. The `invalid_request: client_id is required` failures were caused by a stale `muster` binary at `~/.go/bin/muster` (v0.21.1, predating the workload token-exchange path): the test harness resolves the `muster` binary via `PATH` first (`exec.LookPath` in `getMusterBinaryPath`), so the localmint scenarios ran against the old serve binary instead of the freshly built one. With the current binary first on `PATH`, all six scenarios pass with the CA-pool fix alone; no `client_id` change was needed anywhere.

## Outcome

- All localmint scenarios pass with mcp-oauth v1 (locally and in CI go-build)
- PR #949 merged with the CA-pool migration included (released as v0.22.9)
- `wip/mcp-oauth-v1-migration` branch deleted after landing

Issue #856, opened 2026-06-17
OAuth/OBO/M2M security audit — consolidated findings (3 HIGH, 11 MEDIUM)


QuentinBisson:
## OAuth / OBO / M2M security audit — consolidated findings

Adversarially-verified audit of the OAuth surface across `muster`, `mcp-oauth`, and
`mcp-kubernetes` (token exchange, OBO, M2M workload exchange, K8s impersonation, spec
compliance). Tracking all actionable findings here. Each finding gets the same shape:
**where → impact → fix**. 12 refuted false-positives are listed at the bottom so we
don't re-investigate them.

Counts: **3 HIGH, 11 MEDIUM**, plus LOW/INFO summarized. Findings span three repos;
mcp-oauth and mcp-kubernetes items are cross-linked from here for one place to triage.

---

### H1 — SSO email claim used as `Impersonate-User` without `email_verified` enforcement
- **Repo:** mcp-kubernetes (root cause shared with mcp-oauth SSO path)
- **Where:** `internal/federation/kubeconfig.go:641` & `:732`; `internal/federation/manager.go:866`; `internal/server/oauth_http.go:1078-1085`. mcp-oauth v0.7.1 `server/sso.go:150-168` copies `EmailVerified` but never gates on it. `internal/federation/validation.go:141-189` validates email *format* only.
- **Impact:** A federated IdP emitting an unverified or attacker-controlled `email` claim lets the holder impersonate any K8s user whose RBAC is keyed on that email. Direct privilege escalation (`email: cluster-admin@corp` → become that user).
- **Fix:** Enforce `email_verified == true` before using `email` as the impersonated user, or key impersonation on stable `sub`. Add an SSO-path `allowedClaims` re-check mirroring the existing M2M `sub` re-verification. Reject tokens with `email` present but `email_verified` absent/false.

### H2 — Admin session-detail JSON serializes full signed bearer tokens
- **Repo:** muster
- **Where:** `internal/admin/handlers.go:71-74`; `internal/admin/view.go:24` & `:66-69` (`SessionToken.Raw` has no `json:"-"`); populated at `internal/aggregator/admin_adapter.go:209-244`. HTML path strips the signature via `internal/admin/jwt.go:16-22`, but the JSON branch embeds `*SessionDetail` and emits the full JWT (`header.payload.signature`), including RFC 8693 exchanged access tokens.
[...]

PR #1040, opened 2026-08-20
feat(oauth): send RFC 8707 resource and validate RFC 9207 iss

### LINKED ISSUE #1031: Outbound OAuth flows omit the RFC 8707 resource parameter
## Problem

Muster never sends the RFC 8707 `resource` parameter when it acts as an OAuth client towards
a downstream MCP server. `BuildAuthorizationURL` sets `response_type`, `client_id`,
`redirect_uri`, `state`, `scope`, and the PKCE pair, and nothing else
(`pkg/oauth/client.go:279-302`). The token request adds no `resource` either
(`doTokenRequest`, `pkg/oauth/client.go:227`). `grep -rn '"resource"' pkg/oauth internal/oauth
internal/agent/oauth` returns no hits outside token exchange.

MCP `2026-07-28` §"Resource Parameter Implementation" is unambiguous:

> MCP clients **MUST** implement Resource Indicators for OAuth 2.0 as defined in RFC 8707 […]
> The `resource` parameter **MUST** be included in both authorization requests and token
> requests. […] MCP clients **MUST** send this parameter regardless of whether authorization
> servers support it.

Without it, a token muster obtains for backend A is accepted at backend B whenever both trust
the same issuer. That is the audience-confusion class the parameter exists to prevent.

This is the client-side counterpart to #856 M9, which covers the server side (mintable
audiences not allowlisted). Fixing one does not fix the other.

## Two further client obligations muster does not meet

Verified against the same revision:

- **RFC 9207 `iss` validation.** Clients **MUST** record the issuer from validated AS
  metadata alongside the PKCE verifier, and **MUST** compare a present `iss` on the
  authorization response before redeeming the code, including on error responses. Comparison
  is simple string comparison with no normalisation. Muster records no expected issuer and
  performs no comparison.
- **Credentials keyed by issuer (SEP-2352).** Clients **MUST** key persisted client
  credentials by issuer identifier, **MUST NOT** reuse them across authorization servers, and
  **MUST** re-register when the AS changes.

Two adjacent items are already fine: muster prefers CIMD (`internal/config/defaults.go:56`),
which the revision now favours over Dynamic Client Registration, and DCR is deprecated as of
`2026-07-28` rather than required.

## Proposal

Add `resource` to the authorization request and the token request in `pkg/oauth/client.go`,
using the downstream server's canonical URI (the PRM `resource` value, or the server URL when
`spec.auth.authorizationServer` is used as the opt-out). Apply the same to
`internal/agent/oauth/client.go`. Then add the `iss` record-and-compare step on both, and
key stored credentials by issuer.

Open question: whether to fail closed when the AS does not advertise resource support. The
spec answers it for sending (send regardless), but not for the response side.

## Dual era: what mcp-go v1 changes here

`2026-07-28` is the current spec revision (final, 28 July 2026). Its compatibility matrix has
no soft failures: a legacy client against a modern-only server fails with no fall-forward,
and the reverse also fails. Every muster client and backend is legacy today, so muster must
serve both eras rather than switch.

`mark3labs/mcp-go` v1.0.0-beta.1 (2026-08-12, PR #951) implements `2026-07-28` and is
dual-era: the era is decided per request rather than per build, both eras are served
concurrently on one endpoint, and legacy wire output is pinned unchanged by a conformance
artifact. For this issue specifically:

- **Nothing.** PR #951 states in its own "known gaps" section that authorization hardening is
  untouched: RFC 9207 `iss` validation, `application_type` in Dynamic Client Registration,
  issuer-keyed credentials, and CIMD are an independent workstream in
  `client/transport/oauth*.go`.
- The OAuth flow is transport-level and identical in both eras, so this work is written once
  and is correct before, during, and after the migration.
- Do not wait for the SDK upgrade to start this. No SDK version will deliver it.

Explicitly out of scope for the era adoption in #1035, by that SDK's own admission. This work is neither blocked by nor delivered by it.

QuentinBisson:
### What does this PR do?

Muster acts as an OAuth client towards every OAuth-protected MCP server and sent no RFC 8707 `resource` parameter. It now sends it on both the authorization request and the token request, and validates the RFC 9207 `iss` parameter on the authorization response.

The value always comes from the target's own RFC 9728 metadata (`resource`), for backends and for the `muster agent` login alike, and is sent exactly as declared. A value derived from the configured URL is only the fallback when the metadata omits the field or `spec.auth.authorizationServer` opts out of discovery. This matters twice over: muster binds an issued token's audience to the resource the client sent and then validates tokens against its own `resourceIdentifier`, so a value derived from the endpoint URL yields a token muster refuses; and mcp-go owns token refresh on both paths and puts the declared value on the refresh request verbatim (`client/transport/oauth.go` `refreshToken`), so normalizing it here would bind the initial token and the refreshed token to different audiences. The fallback derivation follows the same constraint: `pkg/oauth.DeriveResourceURI` drops the query and the fragment and changes nothing else, which is what mcp-go does to the same URL (`client/transport/streamable_http.go`), so the two sides agree on the derived value as well.

The `iss` check runs before anything else on the response is read, so a response that cannot be attributed to the expected authorization server is rejected whole, error parameters included. Comparison is a simple string comparison. An absent `iss` is refused when the server advertises `authorization_response_iss_parameter_supported`, and accepted otherwise. Metadata discovery now also applies the RFC 8414 §3.3 identity check, so that comparison cannot target an issuer the authorization server picked for itself.

A token muster obtains for backend A is no longer accepted at backend B when both trust the same issuer, provided the authorization server honors the parameter. An authorization server that rejects an unknown `resource` will fail the login; MCP 2026-07-28 requires sending it regardless.

An authorization server whose metadata reports an issuer other than the URL it was fetched from is now refused. A backend whose `WWW-Authenticate` header points at metadata on another origin no longer has that document followed: discovery falls through to the well-known path, and a login that depended on the foreign pointer fails there. `**Breaking** (Go API)`: `pkg/oauth.Client.BuildAuthorizationURL` takes an `AuthorizationRequest` struct, `ExchangeCode` takes a trailing `resource` argument, and `ResolveResourceIndicator` returns a `ResourceIndicator` struct instead of two errors.

One user-visible reordering in the proxy callback: an error response that carries no valid state renders "missing required parameters" instead of the OAuth denial page, because muster cannot attribute it to a flow.

Closes #1031. Client side only; the server-side counterpart (mintable audiences not allowlisted) is #856 M9.

Refresh requests need nothing from muster: mcp-go already sends the RFC 8707 parameter on them, which is why the declared value must go out unchanged.

SEP-2352 (credentials keyed by issuer) has nothing to key here: muster's `client_id` is a self-hosted CIMD URL, so there are no per-AS credentials, and no DCR path exists. No SDK version delivers this work; mcp-go PR #951 lists authorization hardening in its own known gaps.

### Testing

Table-driven tests cover the four `iss` cases on the proxy callback and the agent flow, the resource parameter on both requests, a declared value that is not canonical going out unchanged, the fallback derivation (percent-encoding, explicit ports, trailing slash, host case, query, fragment, userinfo) against mcp-go's rule, and the RFC 8414 §3.3 rejection. All 182 BDD scenarios pass, plus `go test -race` and `golangci-lint -E gosec -E goconst`.

Live verification on glean is partial: a differential probe shows glean's Dex carries `resource` through the authorization endpoint and connector page unchanged. Glean's Dex does not advertise `authorization_response_iss_parameter_supported`, so the absent-`iss` branch stays permissive there. A full interactive login against a backend still needs a dev image on glean and a browser Dex login.

`docs/explanation/mcp-2026-07-28/05-authorization-hardening.md`: the SEP-2468 bullets and the matching work item are updated, and two coverage limits are recorded — the `iss` check is inert against a server that neither sends the parameter nor advertises it, and refresh requests get their resource indicator from mcp-go. `docs/how-to/connecting-non-rfc9728-mcp-servers.md` describes the value the override sends: the configured MCP server URL with the query and the fragment dropped.

### Review follow-ups

- The declared resource is now bound to the backend that declared it. A `resource_metadata=` pointer is followed only on the backend's own scheme and host (`pkg/oauth.ValidateAdvertisedMetadataURL`), and a `resource` from such a document is used only when it is on that same origin (`pkg/oauth.ValidateAdvertisedResource`). mcp-go applies both checks. The second compares the origin and not the whole URI, because an mcp-oauth backend serving at `<base>/mcp` declares `<base>`. Without this, promoting the declared value to the audience binding let a backend obtain a token for a resource it does not own.
- The RFC 9207 fallback comparison, which runs only when the AS metadata is unreachable, now ignores a trailing slash on either side. It compares against the operator-configured issuer rather than one the server published, and that difference failed a correct response.
- `Manager.HandleCallback` is deleted. It had no production caller, described itself as a testing API, and did not run the `iss` check that the real callback does.
- The post-login redirect helpers move to `internal/oauth/post_login_redirect.go` so `handler.go` stays under the 400-line limit.

- The fallback derivation now matches mcp-go's rule exactly (see above), so the initial token and the refreshed token agree on the derived value too.
- The declared-or-derived decision lives in one place, `pkg/oauth.ResolveResourceIndicator`. The proxy path and the agent path used to disagree about an unusable declared value: one warned and derived, the other failed the login. Both now warn and derive.
- `spec.auth.authorizationServer` fills the resource from the configured URL, so a login on an override-configured backend no longer re-enters RFC 9728 discovery to look for a value that branch can never supply. That branch also shares one discovery client for the process instead of building a fresh one per call, whose metadata cache lived for exactly one fetch.
- `api.AuthInfo` gains `Resource`, so neither field-by-field copy between it and `mcpserver.AuthInfo` drops the value.
- The agent derived its indicator from the normalized server URL, which drops the transport path, while mcp-go derives its own from the endpoint URL and sends that on refresh. The agent now derives from the endpoint URL, so the two agree on that path as well.
- The agent clears the indicator at the start of every `CheckConnection`, so a value declared by one server cannot reach a login against another one, and it reads the well-known document for the resource when the 401 header names the issuer directly and carries no pointer (which is what mcp-go does).
- The well-known candidate list lives in `pkg/oauth.ProtectedResourceMetadataURLs` and is shared by both paths: the path-inserted form first, then the root form. The agent used to probe only the root form.

Note for reviewers: the RFC 8414 §3.3 identity check fails closed and drives every outbound login. Confirm the advertised `issuer` against the configured value on each install, not only on glean.


QuentinBisson:
### Live verification on gazelle

Deployed `b9258f2` as `5.3.1-dev.fix-rfc8707--e-indicator...hb9258f2` on gazelle (both the `agent-platform` and `muster` HelmReleases suspended for the run). Target backend: `miro` (`https://mcp.miro.com`, DCR client, RFC 9728 metadata declares `resource` and `issuer` as `https://mcp.miro.com/`).

Confirmed:

- The authorization request carries `resource=https%3A%2F%2Fmcp.miro.com%2F`, the declared value verbatim, trailing slash kept.
- The resolved value is recorded on the state (`"resource":"https://mcp.miro.com/"`), so the token request sends the same one.
- Miro accepts the parameter: `/authorize` answers 302 to `/consent`, no `invalid_target`.
- A callback carrying a wrong `iss` is refused before the code is read: `Rejecting OAuth callback for server=miro: iss mismatch: response carried "https://attacker.example.com", expected "https://mcp.miro.com/"`.
- 51 of the 53 MCPServer CRs on the install connect, and the RFC 8693 exchange path is unaffected. The two others (`gazelle-mcp-pro`, `miro`) sit in Auth Required, which is their normal state without a stored user token.
- RFC 8414 §3.3 holds for every issuer on this install: gazelle Dex reports `https://dex.gazelle.awsprod.gigantic.io`, and the mcp-oauth backends report an issuer equal to the URL the document was fetched from.

Not covered here:

- The consent click and the token request. Both need a browser.
- The absent-`iss` refusal against a server that advertises the parameter. gazelle Dex and Miro both omit `authorization_response_iss_parameter_supported`, so that branch stays permissive on both. The mcp-oauth backends do advertise it, which makes `gazelle-mcp-pro` the one target on this install able to exercise the strict branch.

One note for the docs limits: `mcp-kubernetes` serves its RFC 9728 document only at the root form, not at the path-inserted form. The shared candidate list in `pkg/oauth.ProtectedResourceMetadataURLs` covers it, but a backend of that shape is a real case rather than a theoretical one.

QuentinBisson:
### Live verification, second round

Re-run on gazelle against the current head `a12d061` (image `5.4.1-dev.fix-rfc8707--e-indicator...ha12d061`), driven through `core_auth_login` on the `miro` backend and the proxy callback.

| Case | Result |
|---|---|
| `resource` on the authorization request | `resource=https%3A%2F%2Fmcp.miro.com%2F`, the declared value verbatim |
| Authorization server accepts it | `/authorize` answers 302 to `/consent` |
| Callback with a wrong `iss` | refused: `iss mismatch: response carried "https://wrong.example.com", expected "https://mcp.miro.com/"` |
| Callback with no `iss`, server does not advertise support | permissive, as documented, and the flow continues to the token endpoint |
| Callback with a matching `iss` | passes the check and continues to the token endpoint |
| Token request | reaches Miro, which answers `invalid_grant - authorization code does not exist`. No `invalid_target`, so the resource indicator does not break the exchange |
| Error response with a valid state | the denial page, attributed to the flow |
| Error response with no valid state | `missing required parameters` |

Still not covered live:

- A completed consent and a real issued token, which needs a browser.
- The strict absent-`iss` refusal. Every backend on this install whose authorization server advertises `authorization_response_iss_parameter_supported` reaches muster through token exchange or token forwarding, so `core_auth_login` never produces a challenge for one. `core_auth_logout` on `gazelle-mcp-pro` releases the server binding only, and the next login reconnects through SSO on the token already held for that issuer. Live coverage of that branch needs a backend that advertises the parameter and for whose issuer the session holds no token.

Local runs on the same head: `go build ./...`, `go vet ./...` and `go test ./...` all pass. The `./muster test` suite reports 185 of 191 scenarios passing, and the same six fail identically on `origin/main` (`b47f1ce`) and also with `--parallel 1`, so they are specific to that machine and not to this branch. CI `go-build`, which runs the same suite, passes on both.

QuentinBisson:
### Completed consent on gazelle

The interactive login now ran to completion on `a12d061`, against the `miro` backend with a real browser consent:

```
Successfully authenticated session=... server=miro
OAuth callback completing - establishing connection for session=... server=miro
Updated MCPServer miro state to connected (health: healthy)
User ... connected to miro with 57 tools, 0 resources, 0 prompts
```

The authorization request carried `resource=https%3A%2F%2Fmcp.miro.com%2F`, the value Miro declares in its RFC 9728 metadata. A tool call with the resulting token succeeds (`x_miro_user_who_am_i` returns the identity), so the token Miro issued for that resource is accepted at the backend. That closes the token-request half of the change on the live path.

One unrelated line appears immediately after the connection, and the 57 tools are reachable regardless:

```
ERROR Failed to register MCP server miro  error="no MCP client available for miro (service state inconsistent)"
subsystem=Aggregator-EventHandler
```

The only case still uncovered live is the strict absent-`iss` refusal, for the reason in the previous comment.

QuentinBisson:
### RFC 9207 iss branches now covered in the scenario suite (`2fc5f72`)

The last uncovered case, an absent `iss` against an authorization server that advertises the parameter, had no live target on gazelle: every backend there whose authorization server advertises it connects through token exchange or forwarding, so no interactive challenge exists for one. It is now covered in the harness instead.

- `mock.OAuthServerConfig.AdvertiseIssParameter` (`advertise_iss_parameter` in a scenario) advertises `authorization_response_iss_parameter_supported`.
- `test_simulate_oauth_callback` accepts `send_iss: true`, which sends the mock's own issuer, and `iss: "<value>"` for a literal. The parameter stays absent by default, which is what the existing scenarios rely on.
- `oauth-iss-required-missing`: a response with no `iss` is refused, and the server stays unauthenticated.
- `oauth-iss-validated`: a foreign `iss` is refused, and the expected issuer completes the flow through to a protected tool call.

Both pass.

### A local-run trap worth knowing

My earlier report of six failing scenarios was wrong about the cause. `getMusterBinaryPath` resolves `exec.LookPath("muster")` before any build output, so a stale `muster` earlier on PATH runs instead of the current build. Here it was a 26 August binary in `~/go/bin`, which is why the failures clustered on recent work (`oauth-duplicate-callback`, the DCR scenarios, transport detection) and why CI, which has no such binary, was green.

With the branch build first on PATH: 191 of 194 scenarios pass, and the three that error under `--parallel 50` all pass at `--parallel 1`, so they are contention, not defects. Two of the 194 are the new scenarios above.

Unrelated nit found on the way: `make test` passes `--parallel 50` while the flag documents a range of 1 to 20.

Issue #1037, opened 2026-08-20
Session tool visibility leaks across users after another user's connector login


QuentinBisson:
## Problem

A user's session gains a tool that a **different** user authenticated for. This breaks the session-scoped tool visibility contract of ADR-006.

Reproduced in `internal/testing/scenarios/session-multi-user-progressive-auth.yaml`. The scenario has two users, `power-user` and `readonly-user`, and two OAuth-protected backends, `server-alpha` and `server-beta`. After `power-user` authenticates with `server-alpha` only, `readonly-user` lists its own tools and gets:

```
{
  "success": true,
  "tool_count": 29,
  "tools": [
    "x_server-alpha_alpha_tool",     <-- leaked: readonly-user never authenticated
    "core_workflow_list",
    ...
  ]
}
```

`readonly-user` has authenticated with neither backend at that point. The step before (`readonly-user-no-tools-initially`) shows an empty backend tool set, so the tool appears only after the *other* user's login.

The listing is not a harness artifact. `test_list_tools_for_user` (`internal/testing/test_tools.go:1065`) drives that user's own MCP client through the `list_tools` meta-tool, so this is what the user's session really returns.

## Why this was not caught

The scenario asserts the absence with `not_contains`, and `not_contains` is not evaluated for `test_*` steps — see the companion issue about `validateTestToolExpectations`. The assertion has been present and silently skipped. Enabling `not_contains` makes the scenario fail at `readonly-user-still-no-tools`.

## What is needed

1. Find where the alpha tool enters the second session. Candidates: the session tool cache (`internal/api/session_tool_cache.go`), the aggregator's `sessionToolFilter`, or the notification path that pushes `list_changed` after a connector login.
2. Fix the leak.
3. Enable `not_contains` for test-tool steps so the scenario guards it from then on.

Scope note: the sibling scenario `session-multi-user-tool-isolation` passes with `not_contains` enabled, so the leak is not universal. The difference between the two scenarios is a good starting point.

QuentinBisson:
Blocked on #1038: the guard assertion is skipped today, so a fix here needs the validator change to stay guarded.

Ben-Home:
[...]

PR #1081, opened 2026-08-26
refactor(testing): evaluate every expectation with one implementation, for both step kinds

### LINKED ISSUE #1038: BDD test-tool steps silently skip not_contains and wait_for_state expectations
## Problem

The BDD framework has two expectation evaluators, and the one used for `test_*` steps checks less than the one used for MCP steps. Expectations a scenario declares are therefore silently dropped, and the step still passes.

- `validateExpectations` (`internal/testing/test_runner.go:792`), used for MCP tool steps: `success`, `error_contains`, `contains`, `not_contains`, `json_path`. `validateExpectationsWithClient` adds `wait_for_state`.
- `validateTestToolExpectations` (`internal/testing/test_runner.go:648`), used for `test_*` steps: `success`, `error_contains`, `contains`, and `json_path` (added in #1036). **`not_contains` and `wait_for_state` are still ignored.**

Nothing warns. A scenario that declares `not_contains` on a test-tool step reports the check in the run output ("JSON path checks: N" style) and passes regardless of the response.

## Impact

Found while verifying #1030: a new step asserting `resources_subscribe: false` passed even with the capability re-enabled. `json_path` is now wired, which uncovered that 8 scenarios were relying on skipped `json_path` checks (all 8 still pass).

`not_contains` is still skipped on 18 steps across 5 scenarios:

- `session-multi-user-progressive-auth.yaml` (5 steps)
- `session-multi-user-tool-isolation.yaml` (5 steps)
- `filter-tools-query-ranking.yaml`, `filter-tools-labels-facet.yaml`, `filter-tools-discovery-summary.yaml`, `list-core-tools-full-detail.yaml`

Enabling `not_contains` today makes `session-multi-user-progressive-auth` fail on a real defect, filed separately.

## Proposal

Collapse the two evaluators into one, so a scenario means the same thing whichever tool kind a step targets. The test-tool path needs a small adapter for its response shape (a plain map rather than an `mcp.CallToolResult`), which is the only real difference between the two.

Land it together with the leak fix, otherwise the suite goes red.

teemow:
Closes #1038 properly. #1078 fixed the *symptom* — `not_contains` was being ignored on `test_*` steps — but left the cause in place: two evaluators, one per response shape, each with its own copy of the checks. That structure has now drifted three times.

| Kind | Implemented on one path in | Silently ignored on the other until |
|---|---|---|
| `json_path` | #1036 | #1036 |
| `wait_for_state` | #1076 | #1076 |
| `not_contains` | (regular path, long-standing) | #1078 |

Each time, scenarios that declared the kind on the wrong step kind passed regardless of the response. That is how the session-isolation leak in #1037 sat behind a green suite: the assertion guarding it existed and did nothing.

## What changed

Both paths now adapt their own response into a `responseView` (`present`, `hasError`, `text`, `errorText`, `json`) and share one `checkExpectations`. Adapting a shape is the only per-shape work left, so a kind cannot be honoured for one step kind and dropped for the other. 278 lines of duplicated checking removed.

Two behaviours are unified rather than kept per-path, since parity is the whole point:

- **`error_contains` is evaluated whenever declared.** The test-tool path used to skip it unless `success: false`. No scenario in the repo pairs it with `success: true` (checked all 187 files).
- **A `success: false` payload fails a step that expected success.** Only the test-tool path checked this.

All 186 scenarios are unaffected by both — no verdict changed.

### A third instance, not yet fired

`status_code` was declared in `TestExpectation` and printed by the reporter, but **never evaluated by either path** — so a scenario asserting an HTTP status asserted nothing. Neither response shape carries one. No scenario in the repo uses it, so nothing was silently passing yet.

It is now rejected at load time, pointing at the supported alternative. Deleting the field instead would have *restored* the silent pass: the YAML decode is not strict, so the key would just be dropped.

## The guard

`TestEveryExpectationKindIsEnforcedOnBothStepKinds` — for each field of `TestExpectation`, assert a violating response is rejected through **both** entry points. A reflection check then fails the build when a field is added without either an enforcement case or an entry in `enforcedElsewhere` naming the test that covers it.

I verified the guard is not itself vacuous by adding a dummy `HeaderContains` kind and confirming it fails:

```
TestExpectation.HeaderContains has no enforcement case.
Every expectation kind the framework accepts must be enforced, or a scenario
declaring it passes vacuously (#1038). Add a case to this test, or record it in
enforcedElsewhere naming the test that covers it.
```

`TestBothStepKindsAgreeOnTheSamePayload` pins the subtler half: the same assertion must reach the same verdict on both shapes, so a kind cannot quietly *mean* different things.

`callTestToolWithWait` now takes a narrow `testToolInvoker` interface instead of the concrete `*TestToolsHandler`, so both `wait_for_state` polling loops are exercised for real (each subtest takes ~1s because the tool genuinely gets re-invoked) rather than assumed.

## Verification

- `go test ./...` — all packages green.
- `muster test` — **186/186**, fresh build (`go build -o ./muster . && go install .`; the harness resolves `muster` from `PATH` first).
- `golangci-lint run ./internal/testing/...` — 0 issues.
- **The #1037 leak guard still bites through the shared path.** I reverted only the `getSessionIDFromContext` fallback, rebuilt, and re-ran: `session-multi-user-progressive-auth` failed at exactly `readonly-user-still-no-tools` (7/8 steps passed). Restored → green. The refactor did not weaken the assertion it was meant to protect.

Not breaking: `internal/testing` only, no change to muster's runtime or API. The one user-visible change is that a scenario declaring `status_code` now fails to load instead of asserting nothing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


teemow:
### CI note: GitHub Actions is not dispatching, so the required GHA checks have not run

All 12 CircleCI checks pass, including `ci/circleci: go-build` (the Go test suite) and `test-envtest`.

The three required **GitHub Actions** checks never produced a result:

- `pre-commit` and `check-values-schema` both came back `startup_failure` within six seconds of the push — the workflows failed to start, not to pass.
- Re-running `pre-commit` and reopening the PR each created a fresh run; all of them are sitting in `queued` with no runner picking them up.
- `semantic-pull-request` produced no run at all.

This is not caused by this branch: the workflow files here are byte-identical to `main` (`git diff origin/main...HEAD -- .github/` is empty), the commit touches nothing under `.github/`, and the same `pre-commit` workflow succeeded on `main` and on a Renovate branch earlier today.

Since the checks cannot report, I ran their substance locally instead:

```
$ pre-commit run --all-files          # same hook set the CI job executes
check for merge conflicts.......................Passed
check that scripts with shebangs are executable.Passed
detect private key..............................Passed
fix end of files................................Passed
mixed line ending...............................Passed
trim trailing whitespace........................Passed
go fmt..........................................Passed
go-mod-tidy.....................................Passed
golangci-lint...................................Passed
go imports......................................Passed
generate + fix + normalize helm/muster/values.schema.json.......Passed
verify JSON Schema..............................Passed
generate + fix + normalize helm/muster-crds/values.schema.json..Passed
verify JSON Schema..............................Passed
Helm Docs.......................................Passed
EXIT=0
```

golangci-lint is the same version CI pins (2.13.1). The two `values.schema.json` generate-and-verify hooks are what `check-values-schema / validate` runs, and they pass with nothing left to regenerate — expected, since this change touches no Helm files. The PR title is a valid conventional commit, and `conventional-pre-commit --strict` already validated the commit message at commit time.

Also verified locally: `go test ./...` green across all packages, and `muster test` 186/186 on a fresh build.

PR #1111, opened 2026-08-28
fix(auth): make the auth handler registry atomic and honor MUSTER_OAUTH_CALLBACK_PORT in agent mode

The127:
### What does this PR do?

Three independent fixes to the CLI-side auth handler lifecycle, all reachable from `muster agent` and the `muster auth` commands.

**1. Get-or-create on the auth handler registry is now atomic.** `ensureAuthHandlerWithOptions` and `cli.ToolExecutor.setupAuthentication` both spelled get-or-create as four steps — `GetAuthHandler`, construct, `Register`, `GetAuthHandler` again — which is three lock acquisitions. `handlerMutex` makes each acquisition atomic but leaves the composite racy, so concurrent callers each built and published their own adapter and every loser was orphaned without `Close()`. New `api.GetOrRegisterAuthHandler` runs the check, the construction and the publish under one write lock, so the factory runs exactly once and every caller observes the same handler. Both sites now go through it.

**2. Closing the registered `AuthAdapter` now clears the registry.** `Register()` publishes an adapter into a process-global registry, but `Close()` only tore down its managers and never unpublished. The global kept pointing at a closed adapter, so a later `api.GetAuthHandler()` returned a husk — and because `AuthAdapter` carries no closed flag, callers silently re-created managers on it instead of failing. New `api.UnregisterAuthHandler` compares and clears under a single write lock, and `Close()` calls it. Closing an adapter that is *not* the registered one stays a no-op.

**3. `muster agent --mcp-server` honors `MUSTER_OAUTH_CALLBACK_PORT`.** `cmd/agent.go` hardcoded its own `DefaultOAuthCallbackPort = 3000` for the `AuthManager` driving that path's OAuth flow. The env var was read in exactly one place — `internal/cli`'s unexported `getCallbackPort` — which that path never consulted. `cli.GetCallbackPort` is now exported as the single resolver and `cmd` delegates to it. The re-auth failure message in `internal/agent/server_mcp.go` no longer names port 3000 either: the number was written into its text, so it stated the default even when the flow had been configured onto another port.

Also adds `api.SwapAuthHandler`, which returns the handler it displaced. `RegisterAuthHandler(nil)` clobbers rather than restores, so tests that reset to nil destroy state belonging to the enclosing test; `Swap` lets them use `t.Cleanup(func() { SwapAuthHandler(prev) })` instead. Adopted in three tests across `cmd` and `internal/cli`.

The `--mcp-server` path no longer registers an adapter at all. Nothing on it reads `api.GetAuthHandler()` — only `ToolExecutor` and the REPL do, and neither runs in that mode — so the `Register()`/deferred-`Close()` pair only overwrote the registry and cleared it again for no reader.

- **`export MUSTER_OAUTH_CALLBACK_PORT=8080` now works for `muster agent --mcp-server`.** Previously it worked for `muster auth login` only: with port 3000 occupied, the agent still bound 3000 and failed — and the port-in-use message interpolated the *env-derived* port, so it could tell you 8080 was in use while the code that actually failed had used 3000. This makes the behavior already documented in `docs/reference/cli/auth.md` true for the agent path.
- The re-auth failure message no longer contradicts itself by naming 3000 when another port was configured.
- No more orphaned adapters, or a login silently running against a closed adapter, when auth setup is entered more than once in a process.
- No config, CRD, or CLI-surface changes. Nothing to migrate.

### How does it look like?

Before — env var set, agent still binds 3000, and the message names the wrong port:

```console
$ export MUSTER_OAUTH_CALLBACK_PORT=8080
$ muster auth login          # binds 8080, works
$ muster agent --mcp-server  # binds 3000, fails
Error: callback port 8080 is already in use. Please free the port and try again
```

After: both bind 8080.

Found while reading the CLI auth paths; no public issue. All three are latent-bug fixes rather than responses to a reported failure — the port one is the only one directly user-visible.

The race in (1) and the use-after-close in (2) are the same underlying shape: a process-global registry mutated through a sequence of individually-locked calls. Both new helpers exist to make the composite operation a single acquisition rather than to add a new capability.

**What this PR does not fix**, so it is not mistaken for more than it is:

- `noSilentRefresh` lives on the shared adapter rather than on the `Login` call, so applying a caller's option remains last-writer-wins. No lock fixes that — ordering the writes does not bind one to its own caller's later `Login()`; only passing the option through `Login()` would. It is unreachable rather than gone: no muster command registers an auth handler twice.
- `AuthAdapter` still has no `closed` flag. Unregistering narrows the window but an in-flight holder can still re-create managers after `Close()`.
- `GetCallbackPort` still falls back silently on an unusable value. A warn there would render nowhere: `pkg/logging` is initialized only by `muster serve`, so every `logging.*` call on a CLI path is already a no-op. Making CLI commands log is separate work.
- `muster agent --mcp-server --silent` accepts the flag and ignores it. That predates this PR — the flag is only read inside `AuthAdapter.Login`, which that path never calls.

Nothing outstanding — all review threads are addressed or answered in the commits. For a second pass:

- **`internal/api/auth_handler.go`.** The factory runs while `handlerMutex` is held, so it must not take that mutex again. That rules out `api.GetXxx`/`RegisterXxx` for the registries `handlerMutex` covers, and, less obviously, closing the handler on a late error — `cli.AuthAdapter.Close` unregisters itself and so takes it. Documented on the function.
- **`DefaultOAuthCallbackPort` is removed** from the `cmd` package (replaced by `agentCallbackPort()`). It was exported; nothing in-tree referenced it.

### Testing

| Test | Pins |
|---|---|
| `TestGetOrRegisterAuthHandler_ConstructsExactlyOnce` | concurrent callers build one adapter, all observe it |
| `TestSwapAuthHandler_ReturnsPrevious` | restore-by-swap for test cleanup |
| `TestAuthAdapter_Close_Unregisters` | `Close()` clears the global |
| `TestAuthAdapter_Close_LeavesOtherHandlerRegistered` | and does not over-reach |
| `TestAgentCallbackPort_HonorsEnvVar` | the agent resolves the port through `cli.GetCallbackPort` |

`TestAgentCallbackPort_HonorsEnvVar` covers the resolver only; observing the argument actually passed to `oauth.NewAuthManager` would need a seam in the construction, and its doc comment says so rather than reading as a call-site guard.

No existing test needed updating: no existing behavior changed shape, only the sequencing behind it. `TestEnsureAuthHandler` now restores the registry it publishes into.

Yes, done. `docs/reference/cli/agent.md` listed only `MUSTER_ENDPOINT`, so the callback port read as belonging to `muster auth login` alone; its environment section now lists `MUSTER_OAUTH_CALLBACK_PORT` too. `docs/reference/cli/auth.md` already documented the variable without scoping it to login, so fix (3) makes the agent match what was already written there. The other two fixes are internal to the API/CLI layers and have no documented surface.

Yes — the callback-port fix, as a user-visible bug fix. The other two have no observable surface outside the process.


The127:
I get the same build/test error on main locally.
Looks like a flaky test setup.

The127:
Rebase fixed the test flakiness.

cmd/agent.go
```diff
@@ -207,19 +213,22 @@ func setupAgentAuthentication(ctx context.Context, client *agent.Client, logger
 		return nil
 	}
 
-	handler := api.GetAuthHandler()
-	if handler == nil {
+	// Get-or-create in one atomic step. Spelling this as GetAuthHandler, then
+	// construct, then Register, then GetAuthHandler again leaves the composite
+	// operation racy even though each call is individually synchronized.
+	handler, err := api.GetOrRegisterAuthHandler(func() (api.AuthHandler, error) {
```
teemow: `ensureAuthHandlerWithOptions` applies the caller's `NoSilentRefresh` to a found handler; this copy doesn't, so `--silent` is silently dropped when an adapter is already registered. Could this just call `ensureAuthHandlerWithOptions(AuthHandlerOptions{NoSilentRefresh: !agentSilentAuth})` instead of duplicating the closure?

cmd/agent.go
```diff
@@ -207,19 +213,22 @@ func setupAgentAuthentication(ctx context.Context, client *agent.Client, logger
...
-		adapter.Register()
-		handler = api.GetAuthHandler()
+		return adapter, nil
+	})
+	if err != nil {
+		logger.Info("Warning: Could not initialize auth adapter: %v", err)
+		return nil
 	}
-
 	if handler == nil {
```
teemow: Dead branch — `GetOrRegisterAuthHandler` never returns `(nil, nil)`.

cmd/agent.go
```diff
@@ -291,7 +300,7 @@ func runMCPServerWithOAuth(ctx context.Context, client *agent.Client, logger *ag
 
 	// First, check if the server requires authentication
 	authManager, err := oauth.NewAuthManager(oauth.AuthManagerConfig{
-		CallbackPort: DefaultOAuthCallbackPort,
+		CallbackPort: agentCallbackPort(),
```
teemow, line 289: The port fix works here, but `adapter.Register()` a few lines up still clobbers any registered handler, and its deferred `Close()` now nils the registry instead of restoring what was displaced — the one site `SwapAuthHandler` would fit. Since nothing reads `api.GetAuthHandler()` in --mcp-server mode (only executor.go/repl.go do), the Register/Close pair might even be deletable. Related: `internal/agent/server_mcp.go:195` still hardcodes “Port 3000” in the re-auth error on this path.

teemow (commented):
Ran the branch locally: build, vet, and the touched packages' tests with `-race` are green; CI is green. Two findings with no diff line to hang them on:

- `internal/cli/executor.go:355-366` still has the exact racy get-or-create this PR removes elsewhere, and `ToolExecutor.Close()` never closes the adapter it registers — so the CHANGELOG's "no longer leaks adapters" doesn't hold on the `muster list/get/call` path yet.
- `docs/reference/cli/agent.md` lists only `MUSTER_ENDPOINT`; the callback-port variable is still documented as auth-login-only.

Inline comments are small and mostly optional; the unconverted `Register()` sites are the ones I'd fix before merge.

cmd/auth_helpers.go
```diff
@@ -42,25 +42,32 @@ func ensureAuthHandler() (api.AuthHandler, error) {
...
 		return nil, fmt.Errorf("failed to initialize authentication: %w", err)
 	}
-	adapter.Register()
 
-	return api.GetAuthHandler(), nil
+	// Apply this caller's option to whichever adapter is now registered, whether
+	// we built it or found it. Returning without this would silently discard the
+	// caller's setting when another caller had already registered an adapter.
+	if adapter, ok := handler.(*cli.AuthAdapter); ok {
+		adapter.SetNoSilentRefresh(opts.NoSilentRefresh)
```
teemow, line 76: This write happens outside the lock, so the last-writer-wins race on `NoSilentRefresh` survives (relocated, not fixed), and `ensureAuthHandler()`'s zero opts reset the flag for everyone. Fine for one-shot CLI processes — but the CHANGELOG shouldn't claim the option race is gone.

internal/api/auth_handler.go
```diff
@@ -116,6 +117,99 @@ func RegisterAuthHandler(h AuthHandler) {
...
+// and registering one via newHandler if none is present yet.
+//
+// The check, the construction, and the publish all happen under a single write
+// lock, so concurrent callers run newHandler exactly once and every caller
+// observes the same handler. Spelling this as separate GetAuthHandler /
+// RegisterAuthHandler calls leaves the composite operation racy even though
+// each individual call is synchronized: racing callers each build their own
+// handler, the last publish wins, and the losers are orphaned without Close().
+//
+// IMPORTANT: newHandler runs while handlerMutex is held, and that mutex guards
```
teemow: Two gaps here: `AuthAdapter.Close()` now also takes `handlerMutex` (via `UnregisterAuthHandler`), so a factory that `Close()`s on a late error self-deadlocks; and the oauth/secret-credentials registries have their own mutexes, so “every handler registry” overstates. Constructing outside the lock and compare-and-publishing (closing the loser) would drop the constraint entirely — and keep `MkdirAll` out of the package-global lock.

internal/cli/auth_adapter.go
```diff
@@ -660,7 +665,17 @@ func (a *AuthAdapter) InvalidateCache(endpoint string) {
...
+//
+// If this adapter is the one currently registered with the API layer, the
+// registration is cleared first. Register() publishes the adapter into a
+// process-global registry that has no expiry of its own, so without this a
+// closed adapter stays reachable through api.GetAuthHandler() -- and because
+// AuthAdapter carries no closed flag, callers would silently re-create managers
+// on it instead of failing. The compare-and-clear is atomic, and a no-op when a
+// different adapter is registered.
 func (a *AuthAdapter) Close() error {
+	api.UnregisterAuthHandler(a)
```
teemow, line 677: This protects future lookups only: with no `closed` flag and `managers` reset to a fresh map below, an in-flight holder silently re-creates a manager (and callback listener) after Close — the exact scenario this comment describes. A `closed bool` checked in `getOrCreateManager` would finish the fix.

internal/cli/auth_adapter.go
```diff
@@ -146,8 +146,13 @@ func (a *AuthAdapter) getOrCreateManager(endpoint string) (*oauth.AuthManager, e
...
 
-// getCallbackPort returns the OAuth callback port from environment or default.
-func getCallbackPort() int {
+// GetCallbackPort returns the OAuth callback port from environment or default.
+//
+// Exported so every OAuth entry point resolves the port the same way. The agent
+// MCP-server path previously hardcoded its own constant and silently ignored
+// MUSTER_OAUTH_CALLBACK_PORT, so `muster auth login` and `muster agent
+// --mcp-server` could disagree about which port to bind.
+func GetCallbackPort() int {
```
teemow, line 155: Invalid values (`70000`, whitespace) still silently fall back to 3000, so the port-in-use error can again name a port the user never set — worth one warn log on the reject path. `internal/agent/oauth.DefaultCallbackPort` also remains an env-blind 3000 fallback for zero-port configs.

CHANGELOG.md
```diff
@@ -77,6 +77,10 @@ All notable changes to this project will be documented in this file.
 
 - The `X-Actor-Token` request header. The actor token is presented once as the RFC 8693 `actor_token` parameter at `/oauth/token`; `/mcp` requests carry only the issued on-behalf-of token as the bearer.
 
+- `muster agent --mcp-server` now honors `MUSTER_OAUTH_CALLBACK_PORT`. The agent's OAuth flow hardcoded port 3000 for its `AuthManager`, while the environment variable was read in exactly one place that this path never consulted — so with 3000 occupied, setting the variable made `muster auth login` work while the agent still bound 3000 and failed, and the port-in-use message interpolated the env-derived port, naming a port the failing code never used. `cli.GetCallbackPort` is now the single resolver for every OAuth entry point, and the `cmd` package's duplicate `DefaultOAuthCallbackPort` constant is removed. ([#1111](https://github.com/giantswarm/muster/pull/1111))
```
teemow: Both new entries are fixes but sit under `### Removed` — move to `### Fixed`.

cmd/agent_callback_port_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"testing"
+
+	"github.com/giantswarm/muster/internal/cli"
+)
+
+// TestAgentCallbackPort_HonorsEnvVar is the regression test for cmd/agent.go:294,
+// which hardcodes the OAuth callback port:
+//
+//	oauth.NewAuthManager(oauth.AuthManagerConfig{
+//	    CallbackPort: cli.DefaultCallbackPort,   // cmd/agent.go:24 — const 3000
```
teemow: This snippet never existed (the removed const was `cmd.DefaultOAuthCallbackPort`), and the cited line numbers were already stale at this PR's HEAD — function names + past tense would survive edits. Also the subtests only pin the wrapper: reverting the `NewAuthManager` call to a literal 3000 stays green; asserting the port actually passed there would pin the real regression.

internal/api/auth_handler_registry_test.go
```diff
@@ -0,0 +1,100 @@
...
+func (s *stubAuthHandler) GetStatusForEndpoint(string) *AuthStatus                 { return nil }
+func (s *stubAuthHandler) InvalidateCache(string)                                  {}
+func (s *stubAuthHandler) Close() error                                            { return nil }
+
+// TestSwapAuthHandler_ReturnsPrevious pins the primitive tests need in order to
+// isolate themselves from the process-global registry.
+//
+// Today the only way to restore the registry is RegisterAuthHandler(nil), which
+// clobbers rather than restores: a test that resets to nil destroys whatever the
+// enclosing test or package had registered. Five test files currently register
```
teemow: I can only find one pre-existing test registering an auth handler (`TestEnsureAuthHandler` in cmd), not five — and `SwapAuthHandler` has no callers outside this file (the new cli tests use Get+Register). Adopt it there, or drop it until it has a user?

internal/cli/auth_adapter_close_test.go
```diff
@@ -0,0 +1,72 @@
...
+// registered handler has to clear the registration.
+func TestAuthAdapter_Close_Unregisters(t *testing.T) {
+	adapter, err := NewAuthAdapterWithConfig(AuthAdapterConfig{
+		TokenStorageDir: t.TempDir(),
+	})
+	if err != nil {
+		t.Fatalf("failed to create adapter: %v", err)
+	}
+
+	prev := api.GetAuthHandler()
```
teemow: This is the Get+Register restore pattern `SwapAuthHandler`'s doc warns about — use the new primitive here. Also `registered` in the second test is never `Close()`d.

teemow:
Nice. Thank you! The comments on code were created by a claude review. Please check if they make sense. Overall this LGTM

teemow (approved):

The127:
A few findings in the review were correct but preexisting issues or would widen widen this PR too much:
- AuthAdapter has no closed flag, so a caller holding one can still re-create managers after Close().
- GetCallbackPort ignores a bad value silently. A warn wouldn't help — CLI commands never init pkg/logging, so every log call outside muster serve is dead.
- noSilentRefresh belongs on Login(), not on the shared adapter.
- muster agent --mcp-server --silent accepts the flag and ignores it.
- docs/reference/cli/auth.md documents a --no-silent flag that doesn't exist.

PR #1119, opened 2026-08-30
fix(filesystem): reject unsafe entity names to close a path-traversal hole

The127:
## Summary

Fixes a **high-severity path-traversal** vulnerability in filesystem (non-Kubernetes) mode. The caller-controlled entity `name` from the `mcpserver`/`workflow` create/get/update/delete tools was joined into the on-disk file path with no validation. A name like `../../../../tmp/evil` escaped the muster config directory, allowing arbitrary `.yaml` **write** (with attacker-influenced content) and **read** (returned via the `get` tool) anywhere the process could reach.

Since these tools are exposed to AI agents, a prompt-injected or malicious agent could escape the config sandbox with a single normal tool call. Kubernetes mode was unaffected — the API server enforces DNS-1123 names; the filesystem store was the unguarded path.

## Fix

Validate the name at `resourceMeta.filePath` in `internal/client/filesystem/store.go` — the single choke point every storage operation (get, create, update, delete, status) flows through. `filePath` now returns `(string, error)` and all call sites reject invalid names before any I/O.

`validateResourceName` enforces **path safety only**: the name must be a single path segment — no path separators (`/`, `\`), no null byte, and not a `.` / `..` reference. This blocks every directory-escape name from the report.

Character policy beyond path safety (e.g. DNS-1123) is intentionally left to Kubernetes admission. Filesystem mode is expected to accept permissive-but-safe names such as `special-chars-workflow_123` (underscore), which the BDD suite documents as valid — enforcing DNS-1123 here would break `workflow-execution-edge-cases`.

## Tests

- `TestCreateMCPServer_RejectsPathTraversal` — exercises the report's PoC names and asserts nothing lands outside the store directory.
- `TestValidateResourceName` — valid names (including underscore/uppercase/space) vs. path-unsafe ones.

Verified **red before green**: with the validator neutered, the traversal cases and all path-unsafe entries fail; they pass with the fix. The `workflow-execution-edge-cases` BDD scenario passes. `go build ./...`, `go vet`, and `goimports` are clean.

### DIFF
--- a/internal/client/filesystem/store.go
+++ b/internal/client/filesystem/store.go
@@ -10,7 +10,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
@@ -30,19 +29,27 @@ func (m resourceMeta) dirPath(basePath string) string {
 	return filepath.Join(basePath, m.dir)
 }
 
-// validateResourceName rejects names that are not safe to turn into a file path.
-// In Kubernetes mode the API server enforces DNS-1123 on metadata.name; the
-// filesystem store is the exposed, unvalidated path, so it must enforce the same
-// constraint before joining the name into a path. DNS-1123 already forbids "/",
-// "\" and empty segments, which closes the path-traversal hole (a name like
-// "../../evil" or "/etc/x" is rejected); the explicit Base check is defense in
-// depth in case the character rule is ever loosened.
+// validateResourceName rejects names that would escape the resource directory
+// when joined into a file path. The name comes straight from a caller-controlled
+// tool argument, and filepath.Join collapses ".." segments, so without this a
+// name like "../../evil" or "/etc/x" writes and reads arbitrary *.yaml files
+// outside the config directory. The name must be a single path segment: no path
+// separators, no null byte, and not a "." / ".." reference. Character policy
+// beyond path safety (e.g. DNS-1123) is intentionally left to the higher layers
+// and to Kubernetes admission; filesystem mode accepts names such as
+// "special-chars-workflow_123" that Kubernetes would reject.
 func validateResourceName(name string) error {
-	if errs := validation.IsDNS1123Subdomain(name); len(errs) > 0 {
-		return fmt.Errorf("invalid name %q: must be a valid DNS-1123 subdomain: %s", name, strings.Join(errs, "; "))
+	if name == "" {
+		return fmt.Errorf("invalid name: must not be empty")
+	}
+	if name == "." || name == ".." {
+		return fmt.Errorf("invalid name %q: must not be a path reference", name)
+	}
+	if strings.ContainsAny(name, `/\`+"\x00") {
+		return fmt.Errorf("invalid name %q: must not contain path separators or null bytes", name)
 	}
 	if filepath.Base(name) != name {
-		return fmt.Errorf("invalid name %q: must not contain path separators", name)
+		return fmt.Errorf("invalid name %q: must be a single path segment", name)
 	}
 	return nil
 }
--- a/internal/client/filesystem/store.go
+++ b/internal/client/filesystem/store.go
@@ -4,11 +4,13 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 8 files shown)


internal/client/filesystem/store.go
```diff
@@ -28,8 +29,36 @@ func (m resourceMeta) dirPath(basePath string) string {
...
+func validateResourceName(name string) error {
+	if name == "" {
+		return fmt.Errorf("invalid name: must not be empty")
+	}
+	if name == "." || name == ".." {
+		return fmt.Errorf("invalid name %q: must not be a path reference", name)
+	}
+	if strings.ContainsAny(name, `/\`+"\x00") {
+		return fmt.Errorf("invalid name %q: must not contain path separators or null bytes", name)
+	}
```
marians: **medium** — this now rejects `\x00` but lets other control characters through, and newline in a name is exploitable via the legacy audit log.

`generateCRDEvent` → `writeLegacyEvent` (`internal/client/filesystem/events.go:374`) writes the name verbatim into a line-oriented log:

```go
fmt.Sprintf("[%s] %s %s/%s: %s - %s (%s)\n", ...)
```

So `core_mcpserver_create` with

```
x\n[2020-01-01T00:00:00Z] MCPServer default/prometheus: MCPServerDeleted - gone (Normal)
```

passes validation, the create succeeds, and two lines land in `events/events.log`. `readLegacyEventsLog` → `parseLegacyEventLine` (`events.go:219`, `:246`) then parses the forged second line into a genuine-looking `EventResult` returned by the event query tools.

Since this validator is now the single choke point for caller-controlled names, rejecting all control runes here fixes that and also avoids creating files with embedded newlines in their names:

```suggestion
	if strings.ContainsAny(name, `/\`) {
		return fmt.Errorf("invalid name %q: must not contain path separators", name)
	}
	for _, r := range name {
		if unicode.IsControl(r) {
			return fmt.Errorf("invalid name %q: must not contain control characters", name)
		}
	}
```

(needs `unicode` in the import block; `unicode.IsControl` covers `\x00`, so the null-byte case stays handled.)

internal/client/filesystem/store_test.go
```diff
@@ -48,11 +48,67 @@ func TestUpdateMCPServerStatus_DoesNotResurrectDeleted(t *testing.T) {
...
+	}
+
+	// Nothing should have been written outside the mcpservers directory.
+	if entries, err := os.ReadDir(base); err == nil {
+		for _, e := range entries {
+			if e.Name() != "mcpservers" {
+				t.Fatalf("unexpected entry written to base dir: %q", e.Name())
+			}
+		}
+	}
```
marians: **low** — this assertion can't detect what it claims, and it's the one part of the test that can escape `t.TempDir()`.

Three separate problems:

1. All six names fail validation, so `mcpservers/` is never created and `os.ReadDir(base)` returns an empty slice — the loop body never executes.
2. The traversal targets (`base/../../escaped.yaml`, `/tmp/muster-traversal-poc.yaml`, `/etc/muster-evil.yaml`) live outside `base`, so scanning `base` wouldn't see them even on a real regression.
3. `if entries, err := os.ReadDir(base); err == nil` silently skips the whole check on error.

And when `validateResourceName` does regress, the test writes into `$TMPDIR`'s parent and attempts `/etc` — outside the temp dir the test owns.

Suggest asserting `os.Stat` on the concrete escaped absolute paths instead, and pointing the traversal names at a sibling `t.TempDir()` rather than `/tmp` and `/etc`, so a regression fails loudly without touching anything outside the test's own directories.

marians (commented):
The core fix is sound and the choke-point claim holds. `resourceMeta.filePath` is the only place in the repo that builds an entity file path from a caller-supplied name (`internal/config/storage.go` is the other on-disk writer and it already runs `sanitizeFilename`), and every store operation — get, list, write, create, update, status, delete — now goes through the validated helper before any I/O. `Patch` routes to `Update` and `Status()`/`SubResource()` route to the same CRUD methods, so there is no bypass. No BDD scenario or internal caller creates entities with separator-bearing or empty names, so no legitimate path regresses. Package tests and `go vet` pass on the branch.

Two inline comments, plus one point that has no line in this diff to hang off:

**`internal/mcpserver/api_adapter.go:1063` — low.** Path safety is enforced only in the store, so `mcpserver_validate` (and the workflow equivalent, `internal/workflow/api_adapter.go:507`) still reports `../../evil` as a valid definition while the subsequent `mcpserver_create` fails. That contradicts the intent stated in that file's own comment at line 1073 ("Rejected at admission time in Kubernetes mode, so `mcpserver_validate` reports it before a caller writes the CR"), and an agent that validates before creating gets contradictory answers. Reusing the same path check in `validateMCPServer` and in workflow validation would keep validate and create in agreement — character policy really can stay with Kubernetes, but path safety is not character policy.

Minor, not worth changing: `listResources` now skips (with an error log) directory entries whose derived name fails validation — in practice only files named exactly `.yaml`/`.yml` (name `""`) or `..yaml` (name `"."`), which previously loaded. And `filepath.Base(name) != name` is dead code on POSIX after the separator check, but it usefully catches `C:foo` on Windows.

marians (approved):
Non-blocking comments added

The127:
Cannot reproduce the go-build test failure locally. 
It looks to be a flaky test.

PR #1120, opened 2026-08-31
chore(lint): silence six gosec false positives with inline nolint

The127:
### What does this PR do?

Adds inline `//nolint:gosec` directives for the six gosec findings that fail `go-test` in this repo, each with the reason next to the code.

architect-orb's `go-test` job runs gosec through golangci-lint with `--no-config`, so this repo's `.golangci.yml` exclusions never load in that job. giantswarm/architect-orb#908 added `--tests=false`, which cleared the test-fixture floods in the other repos in that wave — but none of muster's six, because all of them are in non-test files:

| File | Rule | Why it is a false positive |
| --- | --- | --- |
| `internal/events/types.go` (×4) | G101 | Kubernetes event reason identifiers, not credentials |
| `pkg/oauth/wire.go` | G101 | The RFC 8693 grant-type URN, a protocol constant |
| `internal/testing/mock/oauth_server.go` | G710 | A mock IdP's authorization endpoint must honour the request's `redirect_uri` |

Note `internal/testing/mock` is a helper package compiled into the normal build — a directory name, not a `_test.go` suffix — so `--tests=false` does not reach it.

`.golangci.yml` is deliberately left unchanged: the pre-commit hook runs golangci-lint *with* the config, so that job still needs its existing exclusions. The two mechanisms cover the two jobs.

None. Comment-only change; no behaviour, API, or output is affected.

### How does it look like?

```
ReasonMCPServerTokenForwarded EventReason = "MCPServerTokenForwarded" //nolint:gosec // G101: Kubernetes event reason identifier, not a credential
```

- giantswarm/architect-orb#907 — the `--no-config` problem and the fleet-wide sweep
- giantswarm/architect-orb#908 — the `--tests=false` fix, released in orb v10.1.1

Per the acceptance criteria on #907, this is muster's side: the remaining non-test findings resolved in-repo.

A sanity check that each of the six is genuinely a false positive rather than something worth fixing in code — particularly the G710, which is suppressed on the grounds that the mock is an intentional stand-in for an IdP and honouring `redirect_uri` is the behaviour under test. There is an identical existing precedent at `internal/oauth/handler.go:112`.

### Testing

No tests: comment-only change. Verified locally with `go build ./...`, `go vet`, `goimports` and `go fmt`. I could **not** run golangci-lint locally to confirm the findings clear — this repo targets `go 1.26.0` / toolchain `go1.27.0` and golangci-lint rejects the config when built against an older Go, so CI is the real check on this PR.

No.

- [ ] CHANGELOG.md has been updated (if it exists)

No entry — no user-facing effect, matching the precedent of the previous lint-only changes (#700, #706), neither of which touched the changelog.


marians (approved):

The127:
If you want this, merge it. If not close it. It can be undone quickly and was supposed to be a workaround until a proper solution for https://github.com/giantswarm/architect-orb/issues/907 lands.

Issue #494, opened 2026-03-16
Layer 3: Periodic capability polling for silent redeployment detection


teemow:
## Parent Issue

[#478 - Capability Freshness Architecture](https://github.com/giantswarm/muster/issues/478)

## Objective

Add a background capability poller that periodically re-fetches tool/resource/prompt lists from downstream MCP servers and updates the registry/store if anything changed. This catches scenario 2c from the epic: silent server redeployments where no CR change occurs and no MCP `notifications/tools/list_changed` notification is sent.

## Problem

Layer 2 (push-based notification refresh) handles scenario 2d -- servers that actively send capability-change notifications. However, when a downstream server is silently redeployed (e.g., a Kubernetes pod restart that doesn't trigger a CR update), its tool list may change without any notification being sent. The aggregator's cached tools become stale until the next full reconnection.

## Terminology

This issue distinguishes two server categories based on their connection model, not their authentication mechanism:

- **Non-authenticated servers** (`StatusConnected`): Have a persistent global `ServerInfo.Client`. Tools stored in `ServerInfo.Tools`. One client serves all sessions.
- **Authenticated servers** (`StatusAuthRequired`): Have per-session clients pooled in `SessionConnectionPool`. Tools stored per-session in `CapabilityStore`. Includes manual OAuth login, token forwarding, and token exchange -- the auth mechanism is irrelevant to the polling design.

### New file: `internal/aggregator/capability_poller.go`

A `CapabilityPoller` struct that runs a background goroutine on a configurable interval:

```go
type CapabilityPoller struct {
    aggregator   *AggregatorServer
    interval     time.Duration
    sfGroup      *singleflight.Group  // shared with notification subscriber
}
```

### Polling logic by server type

**Non-authenticated servers** (StatusConnected with persistent `ServerInfo.Client`):

- Call `refreshNonOAuthCapabilities(serverName)` -- this method already exists in `notification_subscriber.go` and does exactly the right thing: re-fetches tools/resources/prompts via the persistent client, deep-compares with `ServerInfo.Tools/Resources/Prompts` using `toolListsEqual`/`resourceListsEqual`/`promptListsEqual`, and updates + notifies only if something changed.
- Singleflight-deduplicated via the existing `notifRefreshGroup` with key `notif-caps/<serverName>`, so a poll that overlaps with a notification-triggered refresh is coalesced.
[...]

Issue #478, opened 2026-03-14
Capability Freshness Architecture: keep tool data and connections always up-to-date


teemow:
## Problem

The aggregator's view of downstream MCP server capabilities goes stale in multiple scenarios:

1. ~~**No cache cleanup for idle sessions** -- SSO tool cache entries live in-memory with no expiry mechanism that works correctly~~ **FIXED by Combined Layer 1+5**
2. ~~**No on-demand re-population** -- once cache entries are gone, there is no mechanism to re-fetch capabilities without re-login~~ **FIXED by Combined Layer 1+5**
3. **Silent redeployments undetected** -- when a downstream server redeploys without a CR change, tool lists remain stale
4. ~~**MCP notifications ignored** -- `notifications/tools/list_changed` from downstream servers is silently dropped because `MCPClient` doesn't expose `OnNotification` and SSO servers have no persistent connection~~ **FIXED by Layer 2** (`OnNotification` added to `MCPClient` interface; notification subscriptions wired for both non-OAuth and SSO servers)
5. **No connection health tracking** -- unavailable servers still list their tools; users see tools, call them, and get errors
6. ~~**Ephemeral client overhead** -- every SSO tool call creates a fresh client (Initialize + CallTool + Close = 3 HTTP round-trips)~~ **FIXED by Layer 6**

## Current State Summary (post Layer 2)

Three stores manage different aspects of SSO server state, all independent:

- **AuthStore** (`SessionAuthStore`): Tracks whether a session may call tools on a server. TTL-based with dual-check (soft `expireAt` + hard cleanup timer).
- **CapabilityStore**: Caches what tools/resources/prompts exist. Keyed by `sessionID + serverName`. 30-day TTL.
- **ConnectionPool** (`SessionConnectionPool`): Caches the live MCP client connection. 1-hour idle timeout with background reaper.

This separation means clearing stale capabilities does not accidentally revoke authentication, and evicting a pooled connection does not lose cached tool metadata. Each store can be refreshed independently.

Two server types with different tool storage:

- **Non-OAuth servers** (StatusConnected): Persistent `MCPClient` in `ServerInfo.Client`, tools stored in `ServerInfo.Tools`. Fetched at registration time. Notification-driven refresh via `OnNotification` re-fetches on capability-change notifications (`notifications/tools/list_changed`, `notifications/resources/list_changed`, `notifications/prompts/list_changed`).
[...]

PR #1164, opened 2026-09-05
fix(aggregator): family tool surface follows what members currently offer

### LINKED ISSUE #1162: Family tool surface keeps tools a member no longer offers: tools/list advertises 40 x_capi_* while every member reports 24
## Problem

When a member of a `spec.family` re-lists **fewer** tools than before (here: mcp-capi 0.3.x hides its 16 mutating tools when running read-only, giantswarm/mcp-capi#194), the aggregated family surface keeps advertising the tools that are gone. Calls to them fail in the metatools layer with `tool not found`, so agents see a tool, try it, and get an error instead of a result.

## Observed (gazelle, 2026-09-05 15:47–16:00Z, muster 5.9.10, agent-platform-mcps 0.8.0 `capi` family with 10 members)

- Every member re-listed 24 tools after the rollout; the connection log has `connected to <mc>-mcp-capi via RFC 8693 token exchange with 24 tools` for gaggle, galaxy, garm, gerbil, glippy, graveler, grizzly, grouse and `via SSO token forwarding with 24 tools` for gazelle. Only glean was in `Failed` (tunnel 504, unrelated).
- `tools/list` for the same user still returned **40** `x_capi_*` names, the 16 hidden ones included. `describe_tool x_capi_pause_cluster`: `Pause cluster reconciliation (available on servers: glean-mcp-capi, glippy-mcp-capi, grouse-mcp-capi)` — glippy and grouse had reported 24 tools (no `capi_pause_cluster`) minutes earlier; glean was Failed and should not contribute at all (ADR-008 / `GetAllTools` skips non-connected servers).
- `call_tool x_capi_pause_cluster` with any `management_cluster` value → `CallTool failed for x_capi_pause_cluster ... error: tool not found: x_capi_pause_cluster` (subsystem metatools), for all ten members.
- The muster pod had restarted at 15:47:28Z, so the stale entries survived a process restart; the per-user re-connections afterwards did not prune them.

## Where to look

`internal/aggregator/registry.go`: `assembleExposedTools` "upserts the familyMappings index by union so per-session contributions add providers without overwriting other sessions' visibility. Deregister is the only path that removes providers from the index." A member that keeps its registration but changes its tool set therefore never loses a provider entry, and whatever feeds the `(available on servers: …)` trailer and the exposed list still sees the old set (per-server `info.Tools` and/or the family index; check both the global and the per-session paths, and whether any tool list is restored from the session store across restarts).

## Proposed solution

Make the family surface follow what members currently offer:

- On every (re-)list of a member's tools, replace that member's contribution to the family buckets instead of unioning: drop it from providers of exposed names it no longer offers, and remove an exposed family tool whose provider list becomes empty.
- Do the same for the per-session contributions, so a user's fresh connection with fewer tools shrinks the list that user sees.
- Emit `tools/list_changed` when the set shrinks, as it does when it grows.
- Do not let a member in `Failed`/`Auth Required` state keep providers in the family index for the exposed listing (it already does not for routing).

## Acceptance criteria

- [x] Integration scenario: a family with two members, one member reconnects advertising a subset of its previous tools → `tools/list` no longer contains the dropped names for that member; a name offered by neither member disappears entirely; `describe_tool` trailers list only current providers.
- [x] A `Failed` member contributes no providers to the family listing.
- [x] Reproduced fix on gazelle: after the mcp-capi 0.3.x rollout the `capi` family lists 24 `x_capi_*` tools, no `tool not found` for a listed name.

## Findings (2026-09-05)

Two mechanisms in `internal/aggregator` kept the stale surface alive; both had to change.

1. **The family routing index only grew.** `assembleExposedTools` upserted `familyMappings` by union on every listing and only `Deregister` removed providers (`registry.go`, `upsertFamilyTool`). A member that stayed registered but stopped offering a tool never lost its provider entry, so `IsFamilyTool` stayed true for names the list no longer carried (or vice versa once another session's pass ran), and `call_tool` ended in `tool not found` / `not available on server`.
2. **The per-session listing served the capability store regardless of the member's state.** `GetAllToolsForSession` read `capabilityStore.Get(session, member)` for every session-auth member. The store outlives the member: a member whose probe fails keeps its pending-auth registry entry (the event handler deliberately skips deregistration of auth-required servers, see `event_handler.go`) and its cached entries, and the store is Valkey, so it survives a muster pod restart -- which is why the 15:47:28Z restart changed nothing. glean (Failed) contributed its pre-rollout 40 tools to the session's list, and through the process-global index to everyone's routing. glippy and grouse were stale for the *listing* session: their `connected ... with 24 tools` lines belonged to other sessions of the same person; the listing session's cached entries had not been refreshed (its reconnect to them had not happened or had failed).

Nothing is restored from the session store into the registry across restarts; the store itself is the persistence.

## Fix

giantswarm/muster#1164:

- `assembleExposedTools` re-syncs the index from every pass: a contributing server's entries are replaced by what it contributed, tools it no longer offers lose it as a provider, a tool no member offers disappears. Servers that did not take part in a pass (other sessions' members, auth-required servers in the global pass) keep their entries, so the cross-session union from #670 still holds.
- A registered server that is not connected, and a session-auth member whose service is down (failed, unreachable, error, stopped, disconnected), contributes nothing to any session's tools, resources or prompts (`ServerInfo.IsDown`).
- A failed SSO (re)connection drops the session's cached capabilities for that server; `list_changed` is pushed when a list shrinks, as it was when it grew.
- Scenario `mcpserver-family-member-relists-subset` covers the acceptance criteria (re-list a subset, name offered by nobody disappears, `describe_tool` trailers list current providers, a Failed member contributes nothing); `test_stop_mock_server` / `test_start_mock_server` were added for the failed-probe phase.

Separate gap found on the way, not fixed here: after a backend restart the session's pooled connection to that member is stale (`session terminated (404), need to re-initialize`) until the session reconnects; tracked under #494 / #478.

## Outcome (2026-09-05)

giantswarm/muster#1164 merged (squash, `ae4cf0cf`), auto-released as **v5.10.1**, chart 5.10.1 picked up on gazelle through the OCIRepository semver range `5.x` (HelmRelease `muster.v79`). **The muster pod restarted once for this rollout, at 2026-09-05T17:36:38Z** (user sessions re-logged in) -- distinct from the 15:47:28Z restart under investigation for the glean tunnel 504s.

The rollout coincided with the tail of the mcp-capi 0.4.0 rollout (kubeconfig export removed, giantswarm/mcp-capi), so it doubled as a live reproduction: minutes before, 13 members reported 23 tools and 10 still 24; by the first request after the restart all 23 connected members re-listed **23** tools for the session (`connected to <mc>-mcp-capi ... with 23 tools`, 17:37:13Z--17:37:15Z), wallaby-mcp-capi was `Failed` (connection refused).

Verified with the CLI context `gazelle` after the restart:

| check | result |
|---|---|
| `muster list tools --context gazelle \| grep -c x_capi_` | **23** (24 on the very first call, while the session's reconnects were still in flight -- the cached entries from before are served until a reconnect replaces them, then the list follows) |
| `describe_tool x_capi_pause_cluster` | `Tool not found` |
| `describe_tool x_capi_get_kubeconfig` (dropped by every member during the rollout) | `Tool not found` |
| `describe_tool x_capi_list_clusters` trailer / `management_cluster` enum | exactly the 23 connected members; the `Failed` wallaby-mcp-capi is not listed; enum equals trailer |
| `x_capi_list_clusters management_cluster=gazelle-mcp-capi` | returns clusters (a listed name is callable) |
| `x_capi_list_clusters management_cluster=wallaby-mcp-capi` | `not available on server "wallaby-mcp-capi" (available: <the 23 members>)` |
| muster log since restart | no `level=ERROR` |

Remaining, out of scope here: a session's pooled connection to a member whose backend restarted stays stale until the session reconnects (`session terminated (404)`), see #494.

teemow:
Fixes #1162.

## Problem

When a `spec.family` member kept its registration but re-listed **fewer** tools (mcp-capi 0.3.x hides its 16 mutating tools when it runs read-only, giantswarm/mcp-capi#194), the aggregated family surface kept advertising the tools that were gone. On gazelle every `<mc>-mcp-capi` member reported 24 tools, `tools/list` still returned 40 `x_capi_*` names, `describe_tool x_capi_pause_cluster` listed glean (Failed), glippy and grouse as providers, and calling it failed with `tool not found`. The stale entries survived a muster pod restart.

## Root cause

Two mechanisms, both in `internal/aggregator`:

1. **The family routing index only grew.** `assembleExposedTools` upserted `familyMappings` by union on every listing and only `Deregister` removed providers. A member that stayed registered but stopped offering a tool never lost its provider entry, so `IsFamilyTool` stayed true and routing/availability disagreed with the list.
2. **The per-session listing served the capability store regardless of the member's state.** `GetAllToolsForSession` read `capabilityStore.Get(session, member)` for every session-auth member. The store outlives the member: a member whose probe fails keeps its pending-auth registry entry (the event handler deliberately skips deregistration of auth-required servers) and its cached entries, and the store is Valkey, so it survives a muster restart. glean was Failed and still contributed its pre-rollout 40 tools to the session's list and, through the process-global index, to everyone's routing.

## Fix

- `assembleExposedTools` now re-syncs the index from each pass (`syncFamilyIndex`): every contributing server's entries are replaced by the family tools it contributed, providers for tools it no longer offers are dropped, and buckets left without providers are deleted. Servers that did not take part in the pass (other sessions' members in a per-session pass, auth-required servers in the global pass) keep their entries, so the cross-session union from #670 still holds (`mcpserver-family-cross-session-corruption` still passes).
- A registered server that is not connected, and a session-auth member whose service is **down** (failed, unreachable, error, stopped, disconnected), contributes an empty set and thereby withdraws its entries. `GetAllToolsForSession` / `ResourcesForSession` / `PromptsForSession` skip down members (`ServerInfo.IsDown`; `StateUnknown` is not down, so directly registered servers as in tests keep working).
- A failed SSO (re)connection drops the session's cached capabilities for that server (`dropSessionCapabilities`), so a member the session cannot reach any more stops being listed for it until a later connection succeeds.
- `tools/resources/prompts list_changed` are now pushed to the person's transport sessions when a list **shrinks**, as they already were when it grew (`refreshSessionCapabilities` and the drop above resolve the OAuth session to its subject through the session tracker).

## Tests

- Unit: `registry_family_sync_test.go` — global and per-session re-list of a subset, cross-session providers kept, down member contributes nothing (tools/resources/prompts), not-connected global member withdraws, `dropSessionCapabilities`, `capabilityMethodsFor`. `go test -race ./internal/aggregator/` green.
- Scenario `mcpserver-family-member-relists-subset` (acceptance criteria): two SSO forwarding members; k8s-b re-lists without `pause_cluster` → `describe_tool` trailer names k8s-a only and routing to k8s-b is rejected naming the current provider; k8s-a drops it too → the name vanishes from `list_tools`, `describe_tool` says not found, calling it fails; k8s-b's backend is stopped and its probe fails (definition, pending-auth entry and cached capabilities survive) → k8s-b contributes nothing; backend back → its cached capabilities count again.
- Test framework: `test_stop_mock_server` / `test_start_mock_server` (same port); protected mock gained `StartOnPort` and force-closes lingering notification streams on `Stop`; both mocks serve on the captured server so a `Stop` racing the `Serve` goroutine cannot dereference a cleared field (the mock unit tests hit that).
- Full suite: `muster test --parallel 20` → 214/214 passed locally.

## Out of scope / follow-up

After a backend restart the session's **pooled connection** to that member is stale (the restarted backend answers `session terminated (404), need to re-initialize`) and nothing re-establishes it until the session reconnects; `core_auth_logout` is not available for SSO servers. That is the redeployment-detection gap tracked in #494 / #478, not the listing bug fixed here; the scenario therefore stops at the listing level after recovery.

## Rollout

gazelle's muster follows the OCIRepository semver range `5.x`; the auto-release from this merge rolls out with an OCIRepository + HelmRelease reconcile. **Note for the glean tunnel 504 investigation:** that rollout restarts the muster pod once more (user sessions re-login); the timestamp will be recorded on #1162 so it can be told apart from the 15:47:28Z restart.


teemow:
Rolled out to gazelle as muster v5.10.1 (chart 5.10.1 via the OCIRepository semver range, HelmRelease muster.v79). **The muster pod restarted once for this at 2026-09-05T17:36:38Z** — for the glean tunnel 504 investigation: this restart is distinct from the 15:47:28Z one. Proof recorded on #1162: 23 x_capi_* tools listed, `describe_tool x_capi_pause_cluster` and `x_capi_get_kubeconfig` not found, provider trailers/enums equal to the 23 connected members with the Failed wallaby-mcp-capi excluded, no ERROR lines in the log.

Issue #1140, opened 2026-09-03
SSO token forwarding: a transport failure to the backend is treated like an auth failure — 5→30 min per-user backoff, reauth_required, call_tool says "tool not found"


teemow:
## What happened

spidertron, muster 5.7.4, 2026-09-03 ~20:10Z. agent-manager 0.2.0 had just become an mcp-oauth resource server (MCPServer `auth: {type: oauth, forwardToken: true}`), but its Cilium egress policy did not yet allow Google's JWKS/userinfo hosts, so the pod could not validate the forwarded id_token and the MCP initialize timed out:

```
20:10:40 INFO  Attempting ID token forwarding for user 10456500... to server agent-manager
20:10:55 WARN  ID token forwarding failed for user 10456500... to server agent-manager: failed to initialize MCP protocol: transport error: failed to send request: Post "http://agent-manager.agent-platform.svc.cluster.local:8080/mcp": ...
20:10:55 WARN  SSO: Connection to agent-manager failed for user 104565002855848627148: ID token forwarding failed: ...
```

The network policy was fixed at ~20:13Z (agent-manager reachable, verified from inside the pod). Every new `muster call --context spidertron x_agent-manager_get_info` session for the next minutes still answered

```
Tool execution failed: tool not found: x_agent-manager_get_info
```

and muster logged nothing about agent-manager at Info level for those sessions. A muster restart at 20:16Z made the very next session forward the token and connect ("ID token forwarding succeeded ... 10 tools").

## Why

`ssoTracker` records the failure per user/server and `HasSSOFailed` gates both `initSSOForSession` and the pool-miss re-init with `ssoBackoffDuration`: 5 minutes after the first failure, doubling per failure up to 30 minutes (`internal/aggregator/server.go`). That is fine for what the code was written for (an upstream refresh chain that returns 401/403), but a *transport* failure gets the same treatment, and the user sees:

- `call_tool` → `tool not found: x_<server>_<tool>` — no hint that the server exists, that SSO forwarding failed, why, or when it is retried;
- the auth status resource reports the server as `reauth_required` (`determineSessionAuthStatus`) — the suggested remedy (re-authenticate) cannot help, the user's token was never the problem;
- the skip is logged at Debug only (`SSO: skipping %s for user %s (failureCount=…, backoff=…)`), so an operator reading Info logs sees a healthy muster that silently ignores the server.

The fix landed for #1135 in 5.7.18 does not touch this path (confirmed by that session).

## What would help

[...]

Issue #493, opened 2026-03-16
Unified MCP Server health checking concept


teemow:
## Unified MCP Server Health Checking Concept

This issue defines the overall health checking architecture for downstream MCP servers in muster. It replaces the previous placeholder description with a concrete, implementable design based on deep analysis of the existing codebase.

## Problem Statement

MCP server health checking in muster currently lacks a unified concept. Health-related logic is scattered across multiple layers with no clear ownership model, leading to confusion about what each layer is responsible for and how they interact.

### Current state: fragmented health tracking

| Layer | Location | What it tracks | Recovery mechanism |
|-------|----------|---------------|-------------------|
| **Service layer** | `internal/services/mcpserver/service.go` | `Start()` / `Initialize()` failures, `CheckHealth()` via Ping | None - CheckHealth() exists but is never called after initial connection |
| **Orchestrator** | `internal/orchestrator/orchestrator.go` | Service states (Failed, Unreachable) for retry eligibility | `retryFailedMCPServers()` with exponential backoff every 30s |
| **Aggregator** | `internal/aggregator/server.go` + `event_handler.go` | Server registration status (Connected, AuthRequired, Unreachable) | None at runtime - registers/deregisters based on state change events |

### Critical gaps

1. **CheckHealth() is orphaned**: The service layer has a working \`CheckHealth()\` method that uses MCP \`Ping()\`, a \`GetHealthCheckInterval()\` returning 30s, and health event generation. But **nothing calls it after initial connection**. Once a server reaches \`StateConnected\`/\`StateRunning\`, it stays there indefinitely regardless of actual health.

2. **No runtime degradation detection**: If a connected MCP server starts returning errors, becomes slow, or its process dies silently, nothing detects this. Tool calls will fail repeatedly, but the server stays registered and its tools stay listed.

3. **Tool visibility is binary**: Tools are either fully visible (server connected) or completely hidden (server not connected). There is no intermediate state for degraded servers where health checks fail intermittently but tool calls still work.

4. **SSO server health is different but already handled**: SSO servers (token forwarding / token exchange) never reach \`StateConnected\` at the service layer (they return 401 at \`Start()\` and get registered as \`StatusAuthRequired\`). Their health is inherently per-session and is already managed via \`callToolWithTokenExchangeRetry\`, \`triggerBackgroundTokenRefresh\`, and the \`ssoTracker\`. **No additional global health tracking is needed for SSO servers.**

[...]

PR #1165, opened 2026-09-05
fix(mcpserver): cap the remote reconnect backoff at 2 minutes and carry the HTTP status and next retry in MCPServerFailed events and the CR status (#1163)

### LINKED ISSUE #1163: Remote-server reconnect backoff doubles to 30 min and hides upstream recovery: glean members stayed Failed/Unreachable 4.5 min longer than the Teleport 504 window (gazelle, 2026-09-05)
## Problem

When a remote MCPServer's initialize fails with a transport error, `internal/services/mcpserver/service.go` counts consecutive failures, marks the server `Unreachable` after `UnreachableThreshold` (3), and schedules the next attempt at `InitialBackoff` (30 s) doubled per failure up to `MaxBackoff` (30 min); the orchestrator only looks every `RetryInterval` (30 s). During a real upstream outage that is fine. When the upstream heals, the member stays `Failed`/`Unreachable` until the next scheduled attempt — after 5 failures that is 8 min, after 7 it is 30 min — although a probe would have shown it healthy.

## Observed (gazelle, muster 5.9.10, 2026-09-05)

The muster pod was replaced at 15:47:33Z (Karpenter spot interruption drained its node). At the same time glean's Teleport app service was answering every new app session with `504 Gateway Timeout`: the agent's connection to the Teleport auth server had gone stale at ~15:44Z, the gRPC keepalive declared it dead at 15:48:30Z (`Failed to serve request … keepalive ping failed to receive ACK within timeout`, `trace.ErrorToCode(ConnectionProblem)` = 504), and the app service recovered somewhere between 15:53:04Z and 15:57:34Z. Muster's schedule for `glean-mcp-capi` and `glean-mcp-prometheus` (identical for both):

| attempt | time (UTC) | result | gap to next attempt |
|---|---|---|---|
| #1 | 15:47:34 | 30 s remote timeout (upstream hung, no response) | 60 s |
| #2 | 15:49:04 | `request failed with status 504: Gateway Timeout` | 90 s |
| #3 | 15:50:34 | 504 → `server unreachable after 3 consecutive failures` | 150 s |
| #4 | 15:53:04 | 504 | 270 s |
| #5 | 15:57:34 | 401 → `Auth Required` (healthy) | — |

`glean-mcp-kubernetes` hit the one moment the app service worked (15:48:34Z) and went `Auth Required` on its second attempt; the other eight family members were never affected. The upstream became healthy at an unknown point inside the 270 s gap, so users saw glean down for up to 4.5 min longer than the outage. One more failure and the gap would have been 8.5 min, then 16.5 min, then 30 min.

The same 504 was also returned for the RFC 8693 token exchange against `dex-glean` (it runs through the same Teleport app service) and was recorded as an SSO failure for the user (`SSO: Connection to glean-mcp-capi failed for user …: token exchange failed … 504`), arming the 5 → 30 min per-user SSO backoff — the transport-vs-auth conflation already tracked in #1140.

## Proposed solution

- Cap the reconnect backoff for remote servers (default ≤ 2 min, configurable) or, keeping the exponential schedule, add a cheap liveness probe (TCP connect or `HEAD`/`GET` on the endpoint) every `RetryInterval` that resets `nextRetryAfter` as soon as the endpoint answers again — the health layer #493 asks for.
- Carry the HTTP status of the failed attempt and the scheduled next attempt in the `MCPServerFailed` event / CR status so an operator can tell "upstream 504" from "connection refused" without reading logs.
- Do not let a 5xx from the exchange endpoint record a per-user SSO failure (scope of #1140; noted here because this incident shows both effects together).

## Acceptance criteria

- [ ] After an upstream outage ends, a remote MCPServer with N ≥ 3 consecutive failures reconnects within `max(RetryInterval, cap)`, cap ≤ 2 min by default, instead of `30 s · 2^(N−1)`.
- [ ] Integration scenario: remote server returns 504 for four attempts and then recovers → the server leaves `Unreachable` within the cap; a 401 after recovery lands in `Auth Required` as today.
- [ ] `MCPServerFailed` event text and `status` carry the HTTP status and the next retry time.
- [ ] Reproduced on gazelle with a member behind a tunnelport RemoteApp: after a forced 504 window the member is back within the cap.

## Related

- giantswarm/tunnelport#110 — the RemoteApp status stayed Ready/TunnelVerified through the same 504 window (sibling issue, same incident)
- #493 (health checking concept), #1140 (transport failure recorded as SSO failure)

teemow:
Fixes #1163. Sibling issue from the same incident: giantswarm/tunnelport#110 (the RemoteApp status stayed Ready through the 504 window; owned separately).

## What happened

On 2026-09-05 a Teleport app service in front of two remote MCPServers answered every new session with `504 Gateway Timeout` for roughly ten minutes. muster counted the 504s as connection failures (correct), marked the servers `Unreachable` after three, and doubled the wait between attempts: 60 s, 90 s, 150 s, 270 s. The upstream recovered somewhere inside the 270 s gap, so users saw the servers down for up to 4.5 minutes after the outage had ended; one more failure would have hidden 8.5 minutes, then 16.5, then 30. Nothing in the event or the CR status said whether the endpoint had answered 504 or refused the connection, or when muster would look again.

## Which of the two proposed fixes, and why

The issue offered a backoff cap or a liveness probe every retry tick that resets the schedule. This PR **caps the backoff**:

- A probe that can tell "recovered" from "still 504" has to interpret HTTP statuses exactly the way the initialize attempt already does (a TCP connect to the tunnel's local end succeeds throughout the outage; a `GET` answers 504 until it answers 401/405/200). It is the initialize minus the JSON-RPC body: not cheaper in code, only in bytes, and it is the health layer #493 describes, which this change deliberately does not build.
- The cap bounds the recovery delay by construction with one number. Cost during a real outage: one initialize `POST` per server every 2 minutes (a hung upstream costs the 30 s remote timeout per attempt), bounded per tick by `MaxConcurrentRetries = 5`.

The schedule still ramps, so short blips still back off: 30 s, 60 s, 120 s, then 120 s for every further failure. After an outage ends, a remote server is retried within the cap plus the orchestrator's 30 s tick, however many attempts the outage cost.

## Changes

**Backoff cap** (`internal/services/mcpserver/service.go`)
- `MaxBackoff` is a variable read from `MUSTER_MCPSERVER_MAX_BACKOFF` (Go duration, default `2m`), next to the existing `MUSTER_MCPSERVER_INITIAL_BACKOFF`. The cap always wins, also over an initial backoff configured above it.

**HTTP status and next retry in the event and the status**
- The service records the HTTP status of a failed attempt (parsed from the transport's error text, the only place it survives to: `request failed with status 504`, `unexpected status code: 502`, `server returned 401 Unauthorized`; `0` for connection refused / DNS / timeout) and the exact wait the next retry was computed from.
- `MCPServerFailed` events read `server unreachable after 4 consecutive failures (endpoint answered HTTP 504, next retry in 2m0s at 2026-09-05T15:57:34Z): <error>`; failures below the threshold emit `connection failure N of 3 before unreachable (...)`; a failure without an HTTP response says `no HTTP response`. The aggregator's event handler no longer emits a second, bare `MCPServerFailed` on the same `Failed` transition, so there is one event per failed attempt and it is the one with the schedule.
- The CR status carries the schedule: `status.consecutiveFailures`, `status.lastAttempt` and `status.nextRetryAfter` have existed in the CRD since the fields were introduced but were never written; the reconciler now mirrors them from the service on every status sync, together with the new `status.lastFailureHTTPStatus` (CRD regenerated with `make generate-crds`; Flux applies it with `crds: CreateReplace`). The fields are cleared once the server connects or answers a 401, and a 401 now also clears the in-memory schedule, so `Auth Required` no longer sits next to a stale retry forecast. `core_service_status` exposes the same keys under `metadata`; `core_mcpserver_get` returns `lastFailureHTTPStatus`; a Failed server whose error is an HTTP 5xx gets the status message `Upstream error (HTTP 5xx) - the server or a gateway in front of it is failing; muster retries with backoff`.

**Test harness** (`internal/testing`)
- New test tool `test_set_mock_server_outage {server, requests, status?}`: a plain or OAuth-protected mock answers its next N requests with the status (default 504) and serves normally afterwards, so the endpoint stays listening (HTTP response, not connection refused). Session-termination `DELETE`s pass through, so N requests fail exactly N attempts.
- Harness instances run with `MUSTER_MCPSERVER_MAX_BACKOFF=3s` (initial 1 s, tick 1 s), so the third and fourth failure wait 3 s instead of 4 s and 8 s and a scenario can assert the exact schedule.
- Mock server `Stop` releases its listener explicitly so `StartOnPort` can rebind at once.

**Scenario** `mcpserver-remote-backoff-cap` (19 steps): four 504s → `unreachable` with `metadata.lastFailureHTTPStatus: 504`; the events name `next retry in 1s`, `2s`, `3s`, `3s` and never `4s`/`8s`; the server is `connected` again with the schedule cleared and its tool answers. Same outage on an OAuth-protected server → `unreachable` the same way, then `auth_required` with no schedule left over. Backend stopped instead → `no HTTP response`, no HTTP status key, reconnects when the backend is back.

Untouched, as agreed: the per-user SSO failure recorded on a token-exchange 5xx (#1140).

## Tests

- `go test ./...` green; new unit tests pin the capped schedule (30/60/120/120 s), the env override, the status extraction, the event texts against an `httptest` 504 endpoint and a refused connection, the 401-clears-schedule rule, the reconciler's status mirroring and the mock outage gate.
- `muster test` full suite: 215/215 with this branch's binary.
- Regression proof: the new scenario run with `muster test` from this branch against an **origin/main `muster serve`** binary in `PATH` fails at `plain-unreachable-with-http-status` (no `lastFailureHTTPStatus`, uncapped schedule).
- `golangci-lint`: 0 issues on the changed packages.

## Rollout verification (gazelle)

Filled in after the auto-release: see the comment below.


teemow:
## Rollout verification on gazelle

**Rollout.** Auto-release cut `v5.10.2`; gazelle's HelmRelease follows the OCIRepository semver `5.x` and applied it at 18:31Z (Helm release `muster.v80`, pod `muster-749ccd8cfd-f7w7r`, image `gsoci.azurecr.io/giantswarm/muster:5.10.2`). The CRD now has `status.lastFailureHTTPStatus` (Flux `crds: CreateReplace`).

**Forced failing window** on the `glean-mcp-prometheus` member (a tokenExchange server; its healthy state after a probe is `Auth Required`, as in the incident):

- 18:33:21Z `RemoteApp mcp-prometheus-glean` scaled to 0 (`spec.replicas: 0`): the member's Service has no endpoint, muster gets `connection refused`.
- 18:33:46Z restart requested via `spec.restartRequestedAt` so the member probes during the window.
- 18:35:21Z the tunnel pod is back (Flux `flux-extras` reverted the patch after 2 minutes, before my own 3-minute restore).
- 18:36:59Z the orchestrator's attempt gets the 401 → `Auth Required`, `consecutiveFailures`/`nextRetryAfter` cleared at 18:37:00Z.

The member was back **98 s after the tunnel returned**, inside cap (2 min) + retry tick (30 s). Throughout the window `status.nextRetryAfter` stayed `lastAttempt + 2m0s` and the events read `... (no HTTP response, next retry in 2m0s at 2026-09-05T18:35:47Z): ...`; with the previous schedule the same number of failures would have put the next attempt 30 minutes out. Events as recorded on the CR:

```
MCPServerFailed  MCPServer glean-mcp-prometheus operation failed: connection failure 1 of 3 before unreachable (no HTTP response, next retry in 30s at 2026-09-05T18:34:16Z): failed to initialize streamable-http MCP client: ... dial tcp 172.31.115.3:8443: connect: connection refused
MCPServerFailed  MCPServer glean-mcp-prometheus operation failed: server unreachable after 3 consecutive failures (no HTTP response, next retry in 2m0s at 2026-09-05T18:35:47Z): ...
```

Also visible in production without any test: `wallaby-mcp-capi` (endpoint not deployed on that side) now reads `consecutiveFailures: 4`, `nextRetryAfter: lastAttempt + 2m`, no `lastFailureHTTPStatus` — a retry every two minutes instead of every thirty.

**What the gazelle run did not cover.** The tunnel returns `connection refused`, not a `504`, so `lastFailureHTTPStatus` stayed absent here; the `endpoint answered HTTP 504` path is covered by the unit tests and the `mcpserver-remote-backoff-cap` scenario (four 504s from a mock behind an outage gate), not by a live 504 window.

**Found on the way, pre-existing:** a pending `spec.restartRequestedAt` on a down endpoint is re-run by the reconciler on every service state change — 131 restart attempts between 18:33:46Z and 18:34:39Z while `nextRetryAfter` said "2 minutes"; the loop stopped when the field was removed. Filed as #1166; the orchestrator's schedule (this PR) was honoured throughout, the reconciler's restart path was not.

PR #1186, opened 2026-09-09
fix(mcpserver): recover the MCP session after a streamable-http backend is redeployed

### LINKED ISSUE #999: Downstream streamable-HTTP session lost after server redeploy: muster reports healthy while every tool call fails with "Missing session ID"
## Problem

When a downstream non-OAuth `streamable-http` MCP server is redeployed, muster keeps its now-dead MCP session and **every** tool call to that server fails indefinitely, while `core_mcpserver_list` / `core_service_status` continue to report the server as `connected` / `healthy`. There is no automatic recovery — only a manual `core_service_restart` fixes it.

Observed in production on the gazelle installation: all 42 `x_pd_*` (PagerDuty) tools were unusable for ~5 hours with no signal anywhere that anything was wrong.

## Reproduction / evidence

Server definition (`core_mcpserver_list`):

```json
{
  "name": "gazelle-mcp-pagerduty",
  "type": "streamable-http",
  "auth": {"type": "none"},
  "autoStart": true,
  "toolPrefix": "pd",
  "url": "http://mcp-pagerduty.mcp-pagerduty.svc.cluster.local:8080/mcp",
  "sessionStatus": "connected",
  "state": "Connected"
}
```

Every tool call failed identically:

```
x_pd_list_oncalls  -> Tool execution failed: failed to call tool:
                      invalid request: Bad Request: Missing session ID
x_pd_list_services -> (same)
x_pd_list_teams    -> (same)
x_pd_get_user_data -> (same)
```

Meanwhile `core_service_status` reported the server as fully healthy:

```json
{
  "name": "gazelle-mcp-pagerduty",
  "state": "connected",
  "health": "healthy",
  "metadata": {
    "clientReady": true,
    "consecutiveFailures": 0,
    "lastAttempt": "2026-07-30T03:44:20.140679556Z"
  }
}
```

The timestamps show the session outlived the pod it belonged to:

| Fact | Timestamp |
|---|---|
| muster `lastAttempt` (last connection attempt) | `2026-07-30T03:44:20Z` |
| downstream pod `mcp-pagerduty-64b9865cf5-pmg94` created | `2026-07-30T09:00:20Z` (0 restarts) |

The pod was replaced ~5h15m *after* muster last connected, so muster's `Mcp-Session-Id` referred to a process that no longer existed. Downstream pod logs confirm muster's requests arrive without a session the server recognises — a fresh transport is created and the request is then rejected:

```
INFO:mcp.server.streamable_http_manager:Created new transport with session ID: 32e3c848c6bc4edcba4e6212c1e6b4fc
INFO:     100.64.63.23:47182 - "POST /mcp HTTP/1.1" 400 Bad Request
INFO:mcp.server.streamable_http_manager:Created new transport with session ID: cc00d0ce8d604d728f649429dfe4f158
INFO:     100.64.63.23:47192 - "POST /mcp HTTP/1.1" 400 Bad Request
INFO:mcp.server.streamable_http_manager:Created new transport with session ID: 2a3f2eba4a0e407fa2e1734770837ed8
INFO:     100.64.63.23:51514 - "POST /mcp HTTP/1.1" 400 Bad Request
```

Two things worth calling out:

1. **`consecutiveFailures` stayed at `0`** across four-plus failed tool calls — tool-call failures against a connected non-OAuth server are not recorded anywhere.
2. `gazelle-mcp-runbooks` (same `auth: none`, same `streamable-http` pattern) kept working throughout — it simply had not been redeployed. So this is specifically triggered by the downstream redeploy, not by anything account- or config-specific.

**Recovery**: `core_service_restart` on `gazelle-mcp-pagerduty` fixed it immediately — the very next `x_pd_list_oncalls` returned 68 records.

## Why the current health model misses it

This is a concrete production instance of the gaps already documented in #493:

> **CheckHealth() is orphaned**: The service layer has a working `CheckHealth()` method that uses MCP `Ping()` [...] But **nothing calls it after initial connection**. Once a server reaches `StateConnected`/`StateRunning`, it stays there indefinitely regardless of actual health.

> **No tool call failure tracking**: When tool calls fail against a connected server, the failures are not recorded anywhere.

and of #478 scenario **2c (server redeployed, no CR change, no notification)**, which for non-OAuth servers is listed as *Not fixed*.

One detail specific to `streamable-http` that is worth folding into that design: **the failure mode here is MCP session loss, not connectivity loss.** The Service DNS name and port stay valid, TCP connects fine, and the new pod answers HTTP — so any check that only probes reachability (TCP dial, HTTP GET, k8s readiness) will report healthy. Recovery requires re-running `initialize` to obtain a new `Mcp-Session-Id`.

The good news is that #493's Phase 1 as specified would catch this: MCP `Ping()` is itself a session-bound request, so it would have failed with the same 400 and tripped the threshold within ~30s. This report is mainly evidence that Phase 1 + Phase 2 are worth prioritising, plus one cheap targeted addition below.

## Proposed solution

Primary — as already planned in #493:

1. **Phase 1** (orchestrator periodically calls `CheckHealth()` on connected servers) would detect this within one health-check interval and trigger `Restart()`, which re-runs `initialize`.
2. **Phase 2** (per-server tool-call failure tracking) would catch it even faster here, since every call was failing.

Additional, cheap and targeted:

3. **Treat a lost downstream session as its own signal.** When a `streamable-http` downstream returns HTTP `400` with `Missing session ID` (or `404` for an unknown/expired session ID), that is unambiguous: our session is gone but the server is up. Rather than waiting for a failure threshold, re-run `initialize` once to obtain a fresh session and retry the call — the same shape as the existing `callToolWithTokenExchangeRetry` 401 → re-exchange → retry-once path, but keyed on session loss instead of auth expiry. This makes the common redeploy case self-healing and invisible to users.
4. **Don't report `health: healthy` / `clientReady: true` on the basis of a stale successful connection.** At minimum, surface the age of `lastAttempt` so `core_mcpserver_list` output is not actively misleading during an outage.

## Acceptance criteria

- [ ] After a downstream `streamable-http` server is redeployed, tool calls through muster recover without operator intervention
- [ ] A `400 Missing session ID` / `404 unknown session` response from a downstream server triggers a single re-`initialize` + retry of the failed call
- [ ] Tool-call failures against a connected non-OAuth server increment a per-server failure counter (reset on success)
- [ ] While a downstream session is dead, `core_mcpserver_list` / `core_service_status` no longer report `state: connected` + `health: healthy` + `consecutiveFailures: 0`
- [ ] Test coverage for the redeploy case: establish a session, replace the downstream server so the old session ID is invalid, assert the next tool call succeeds

## Related

- #478 — Capability Freshness Architecture (scenario 2c, silent redeployments; Layer 4 health tracking)
- #493 — Unified MCP Server health checking concept (gaps 1 and 5; Phases 1 and 2)

fiunchinho:
## Problem

When a stateful `streamable-http` backend is redeployed, every tool call through muster fails until an operator runs `core_service_restart`, while `core_service_status` keeps reporting `connected` / `healthy`. On gazelle this took all 42 `x_pd_*` tools away for hours (#999, and again today).

## Root cause

The mechanism is one step more precise than #999 describes:

1. The redeployed pod does not know muster's `Mcp-Session-Id` and answers HTTP 404.
2. mcp-go clears the session id from the transport and returns the typed `transport.ErrSessionTerminated`. Its own comment says "the caller must re-initialize".
3. Nothing in muster handled that error. From then on every request went out with **no** session id, which a stateful server rejects too (the Python SDK: `400 Bad Request: Missing session ID`, mcp-go: `404 Invalid session ID`).
4. The standalone GET notification listener usually meets the 404 first, so by the time a tool call fails the session is already gone and the error is no longer the typed one.

## Fix

`baseMCPClient` (shared by every transport) gains a session-recovery path, `client_session_recovery.go`:

- A lost session is recognised by the typed error **or** by a transport that had a session at initialize and has none now.
- The dead mcp-go client is closed, the transport's `connectLocked` performs a fresh handshake, and the failed operation is retried once.
- Concurrent callers that failed on the same dead session share one handshake (generation counter). A handshake that fails leaves the client disconnected with `reconnectPending` set, so the next operation tries again instead of failing until a restart. An explicit `Close` never reconnects.
- Every operation goes through it: tool calls, tool/resource/prompt listings, resource reads, prompts and the health ping. So the notification-driven refresh and a future health probe heal a dead session as well.
- `StreamableHTTPClient` and `DynamicAuthClient` (OAuth) are wired; stdio and SSE have no session to lose and are unchanged.

## Tests

- Unit tests with a fake mcp-go client for each branch: typed error, session cleared by the listener, unrelated errors left alone, stateless server, failed handshake retried by the next call, 16 concurrent callers share one handshake, no reconnect after `Close`, ping recovers.
- `TestStreamableHTTPClient_RecoversSessionAfterBackendRedeploy` runs the real mcp-go stack: a stateful server is swapped for a fresh process behind the same URL and the next call succeeds.
- BDD scenario `mcpserver-session-recovery-after-redeploy`: stop and start the mock backend, the next `x_..._probe` call succeeds and the service stays connected. Passes locally.
- `go test ./... -race` passes except `TestFindAvailablePort_*` in `internal/testing`, which fails identically on unmodified `main` on this machine.

`make lint` could not run locally: the installed golangci-lint is built with Go 1.26 and refuses the repo's `toolchain go1.27.1`. `gofmt` and `go vet` are clean.

## Follow-ups

- #493 Phase 1 (periodic health probe with restart) in a separate PR.
- The acceptance item "failure counter per server / status not `healthy` while dead" of #999 belongs to that PR.

Closes #999


fiunchinho:
Full BDD suite (`muster test --parallel 30`) on this branch: 230/230 scenarios pass locally.

paurosello (approved):

PR #1187, opened 2026-09-09
feat(orchestrator): probe connected MCPServers and restart the ones that stop answering

fiunchinho:
## Problem

Once an MCPServer reaches `connected` / `running` it stays there whatever happens to the connection afterwards. `Service.CheckHealth()` (an MCP ping) exists, but nothing calls it after the initial connect, so a backend that is redeployed, hangs or dies keeps its tools listed and every call fails until an operator runs `core_service_restart`. This is gap 1 of #493 and the "reports healthy while every call fails" half of #999; on gazelle it hid the PagerDuty outage for hours.

## Change

**Phase 1 of #493, in the shape proposed in review:** the orchestrator probes, the service fails itself, the existing reconnect loop restarts.

- `internal/orchestrator/health_check.go`: the maintenance loop gains a second ticker, `HealthCheckInterval` (30 s, `MUSTER_ORCHESTRATOR_HEALTH_CHECK_INTERVAL`). On each tick every connected/running MCPServer service is probed concurrently. That is all the orchestrator does; no restart, no slot, no in-flight map.
- `Service.CheckHealth()` owns the outcome:
  - Servers with no shared client are not probed: `forwardToken` / `tokenExchange` servers and per-session OAuth servers, which a login syncs to Connected while `s.client` is nil.
  - Each probe is bounded by the server's own `spec.timeout` (default 30 s), the same budget initialize gets.
  - The probe is an MCP ping; on a connection that negotiated protocol 2026-07-28, where mcp-go answers ping locally without I/O, it is `tools/list` instead.
  - A JSON-RPC error reply (method not found, ...) is an answer from a live server and is not counted. A probe cancelled by the caller (shutdown) is not counted. A probe against a server that is no longer Running/Connected is not counted.
  - At `HealthCheckFailureThreshold` (3) failures in a row the client is closed and the server moves to `Failed` / `unhealthy` with `nextRetryAfter` due now. The reconnect schedule's own `consecutiveFailures` is left alone so the backoff of the attempts that follow is not inflated. `MCPServerHealthCheckFailed` is emitted once, naming the count and the error.
  - From there the pieces that already exist do the rest: the aggregator deregisters the tools on Failed, the reconciler mirrors `Failed` / `lastError` / `nextRetryAfter` into the CR status, and `retryFailedMCPServers` restarts the server on its next tick with its `MaxConcurrentRetries` bound (`MCPServerRecoveryStarted`).
  - A reconnect that follows failed probes is scheduled again whatever error its start hits, transient or not (`reconnectAfterProbe`, cleared by the next attempt that reaches the endpoint). Before, a 404 / `ErrLegacySSEServer`, an x509 error or an HTML login page during the outage left the server `Failed` with no schedule, never retried.
- The aggregator event handler no longer emits its own `MCPServerHealthCheckFailed` on Running+Unhealthy; the service's gated emission is the one that stays.
- `core_service_status` exposes the running count as `metadata.consecutiveHealthCheckFailures`.
- Docs: `MUSTER_ORCHESTRATOR_HEALTH_CHECK_INTERVAL` is documented next to its three siblings in `docs/how-to/mcp-server-management.md`.
- The BDD harness runs instances with `MUSTER_ORCHESTRATOR_HEALTH_CHECK_INTERVAL=1s`; the mock servers' outage gate lets pings through unless armed with `pings: true`.

## Interplay with #1186

The session-recovery fix (#1186, merged) makes the ping itself heal a lost session, so a redeploy is usually repaired by the probe or the next call without any restart. This PR is the safety net for everything the client cannot heal: a hung backend, a dead process, a gateway that fails every request.

## Tests

- `internal/services/mcpserver/health_check_test.go`: the threshold closes the client and produces `Failed` with a reconnect due now and one event, the schedule's failure count untouched; a passing probe resets the count; no shared client and session-level auth are not probed; a non-running server is not counted; a JSON-RPC error reply and a cancelled probe are not counted, the probe's own deadline is; a 2026-07-28 connection is probed with `tools/list`; the probe deadline follows `spec.timeout`; `Stop` resets the count; a reconnect after failed probes is scheduled even when initialize hits a 404 (real `httptest` server), while a first connect against the same 404 stays unscheduled as before.
- `internal/orchestrator/health_check_test.go`: connected and running servers are probed and nothing is restarted by the probe; seven inactive states are not probed; services without `HealthChecker` are skipped; nothing runs after shutdown; the health ticker drives probes; a service that is `Failed` with `nextRetryAfter` now is restarted by the retry loop.
- BDD scenarios: the three from review (`mcpserver-health-probe-spares-session-auth-server`, `mcpserver-health-probe-spares-per-session-oauth-server`, `mcpserver-health-probe-restart-failure-is-retried`) pass, as do `mcpserver-health-probe-restarts-dead-backend`, `mcpserver-session-recovery-after-redeploy`, `mcpserver-restart-request-endpoint-down` and `mcpserver-remote-backoff-cap`.
- `go test -race` on `internal/orchestrator`, `internal/services/...`, `internal/aggregator`, `internal/testing/mock`; `golangci-lint` with `gosec` and `goconst`: clean.

## Not in this PR

- #493 Phase 2 (tool-call failure counter in the aggregator) and Phase 3 (health column in `core_mcpserver_list` / CLI).
- Mirroring `consecutiveHealthCheckFailures` into the CR status (needs a CRD field).
- Chart values for the four `MUSTER_*` tuning variables; today none of them can be set from the chart.
- A per-service lifecycle mutex for Start/Stop/Restart. This PR adds no new caller of those, but the reconciler and the retry loop can still overlap as before.
- #478 Layer 3 (capability polling) is PR #496, which needs a rebase.

Refs #493, #999, #478


fiunchinho:
Full BDD suite (`muster test --parallel 30`) on this branch with the 1 s probe interval: 230/230 scenarios pass locally.

internal/orchestrator/health_check.go
```diff
@@ -0,0 +1,144 @@
...
+// within it counts as a failed probe.
+const HealthCheckTimeout = 10 * time.Second
+
+// checkConnectedServersHealth probes every connected or running MCPServer
+// service once, concurrently, and restarts the ones that report unhealthy.
+// A service whose previous probe or restart is still running is skipped.
+func (o *Orchestrator) checkConnectedServersHealth() {
+	for _, svc := range o.registry.GetByType(services.TypeMCPServer) {
+		checker, ok := svc.(services.HealthChecker)
+		if !ok || !isActiveState(svc.GetState()) {
```
teemow, line 41: Every Connected service is probed, but servers with session-level auth (`forwardToken` / `tokenExchange`) and per-session OAuth servers are set Connected by `notifyMCPServerConnected` → `api.UpdateMCPServerState` after a login while `s.client` is nil (the anonymous probe was discarded, or `Start` returned on the 401). `CheckHealth` then hits `client == nil`, `recordHealthCheckFailure` counts it (`IsRunning()` is true), and on the third tick the server is declared unhealthy, `MCPServerHealthCheckFailed` fires and the loop restarts it: Stop → Start → 401 / `discardAnonymousProbe` → `auth_required`, `MCPServerRecoveryFailed`. That undoes the login's state sync 90 s after every login in production and re-arms on the next session connect.

Reproduced with this branch's binary (`oauth-sso-token-forwarding` and `oauth-token-loss-terminal-auth-required` extended past 3 s: connected → "failed 3 consecutive health checks, restarting" → auth_required), and it is already visible in `mcpserver-family-member-relists-subset`'s instance log ("Health check failure 1 of 3: MCP client not available" for k8s-a and k8s-b — that scenario ends 20 ms before the third tick, which is why CI is green). The count also carries across these `UpdateMCPServerState` transitions, since only Start, Stop and a passing probe reset it.

Suggest not probing services that have no shared client (`UsesSessionAuth()` / `!IsClientReady()`) instead of counting a nil client as a probe failure. Scenarios: `mcpserver-health-probe-spares-session-auth-server`, `mcpserver-health-probe-spares-per-session-oauth-server`.

internal/services/mcpserver/health_check_test.go
```diff
@@ -0,0 +1,198 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, []services.HealthStatus{services.HealthUnhealthy, services.HealthHealthy}, rec.published())
+	for i := 0; i < HealthCheckFailureThreshold; i++ {
+		pinger.fail(errors.New("connection refused"))
+		_, _ = svc.CheckHealth(ctx)
+	}
+	assert.Equal(t, 2, cm.count("MCPServerHealthCheckFailed"))
+}
+
+func TestCheckHealth_MissingClientCountsAsFailure(t *testing.T) {
```
teemow: This test codifies the nil-client counting from the comment on `health_check.go:43`; with the fix it should assert the opposite — a missing shared client is not a probe failure.

internal/orchestrator/health_check.go
```diff
@@ -0,0 +1,144 @@
...
+
+	logging.Warn("Orchestrator", "MCPServer %s failed %d consecutive health checks, restarting: %v",
+		svc.GetName(), mcpserver.HealthCheckFailureThreshold, err)
+
+	if err := svc.Restart(o.ctx); err != nil {
+		if api.IsAuthRequiredError(err) {
+			// Pending auth registration happens in the auth-required hook inside Start.
+			return
+		}
+		// Start scheduled its own reconnect attempts; the retry loop takes it from here.
```
teemow: Not for every error: `Start` schedules `nextRetryAfter` only for `isRemoteServer() && isTransientConnectivityError(err)`; anything else sets Failed with no schedule and `shouldAttemptRetry` skips it forever ("No retry backoff set … skipping automatic retry"). So a health-triggered restart whose initialize hits a non-transient error strands the server until someone restarts it by hand. mcp-go maps any 4xx on the initialize POST to `ErrLegacySSEServer` ("server returned 4xx for initialize POST, likely a legacy SSE server"), which matches no transient pattern — a route answering 404/403/429 during the rollout that failed the pings does it, as do x509 errors and an HTML login page. Before the probe existed the connection stayed Connected and served again when the backend returned.

Scenario: `mcpserver-health-probe-restart-failure-is-retried` (uses the `pings: true` outage flag added to the mock in 3799f348).

internal/orchestrator/health_check.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if state := svc.GetState(); !isActiveState(state) {
+		logging.Info("Orchestrator", "MCPServer %s is %s since its failed health check, not restarting it", svc.GetName(), state)
+		return
+	}
+
+	logging.Warn("Orchestrator", "MCPServer %s failed %d consecutive health checks, restarting: %v",
+		svc.GetName(), mcpserver.HealthCheckFailureThreshold, err)
+
+	if err := svc.Restart(o.ctx); err != nil {
```
teemow: `Start`/`Stop`/`Restart` have no mutual exclusion, and this adds an autonomous caller that fires exactly when operators and the reconciler also restart. Two consequences:

1. Two `Start`s both pass the `IsRunning()` guard (Starting is not running) and `createAndInitializeClient` assigns `s.client` unconditionally, so the second client overwrites the first without closing it — a leaked GET listener goroutine or stdio subprocess — and the second `UpdateState(Connected)` is a no-op transition, so the aggregator keeps serving the first client while `CheckHealth` pings the second (a config change in between is silently lost).
2. The state re-check on line 86 and this `Restart` are not atomic. A `RemoveService` (CR deleted) or a filesystem-mode `core_service_stop` landing inside Restart's Stop → 200 ms grace → Start window is undone; for `RemoveService` the Start completes on an unregistered service (`publishStateChangeEvent` drops the event because `registry.Get` fails) whose live client nothing can reach until process exit.

A per-service lifecycle mutex, or a generation counter re-checked after the grace sleep, covers both.

internal/orchestrator/health_check.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	slots := o.healthRestartSlotsChan()
+	select {
+	case slots <- struct{}{}:
+		defer func() { <-slots }()
+	case <-o.ctx.Done():
+		return
+	}
+	if state := svc.GetState(); !isActiveState(state) {
```
teemow: After waiting for a slot only the state is re-checked, not the health. A slot is held for the whole `Restart` (Close DELETE 5 s + grace + up to 30 s init), and an operator or `restartRequestedAt` restart of a waiting server leaves it Connected/Healthy with the count reset by `Start` — this goroutine then restarts the fresh connection a second time. `svc.GetHealth() == HealthUnhealthy` (Unhealthy is set only at the threshold or by a failed Start), or one re-probe after acquiring the slot, would skip it.

internal/orchestrator/health_check.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	probeCtx, cancel := context.WithTimeout(o.ctx, HealthCheckTimeout)
+	health, err := checker.CheckHealth(probeCtx)
+	cancel()
+	if health != services.HealthUnhealthy {
+		return
+	}
+
+	slots := o.healthRestartSlotsChan()
```
teemow: This semaphore and the truncation to `MaxConcurrentRetries` in `attemptReconnectFailedServers` don't see each other: health restarts that fail land in Failed with `nextRetryAfter` = InitialBackoff, and the next retry tick starts five more while up to five health goroutines still hold slots, so ten initializes hit the same upstream. Meanwhile the queued unhealthy servers sit Connected+Unhealthy with their tools withdrawn for the whole drain (~35 s per five servers under a hung upstream), while probes and `Orchestrator.Start` are unbounded anyway. One bounded restart executor fed by both paths — or the shape in the summary — removes both.

internal/orchestrator/health_check.go
```diff
@@ -0,0 +1,144 @@
...
+// every server behind it rebuild its client in the same instant, and it is
+// skipped when the service left Connected/Running while the probe ran: an
+// operator's core_service_stop or a reconciler restart got there first and a
+// restart now would undo the stop.
+func (o *Orchestrator) probeAndRecover(svc services.Service, checker services.HealthChecker) {
+	if o.ctx.Err() != nil {
+		return
+	}
+
+	probeCtx, cancel := context.WithTimeout(o.ctx, HealthCheckTimeout)
```
teemow: `probeCtx` derives from `o.ctx`, so on shutdown an in-flight ping returns `context.Canceled` and is counted (services are never stopped on shutdown, so `IsRunning()` stays true); a third one flips a server unhealthy and writes an event during shutdown. The `select` below also picks between an always-ready slot and the closed `ctx.Done()` at random. Skip counting when `o.ctx.Err() != nil` or the error is a context error, and check ctx before the select.

internal/orchestrator/health_check.go
```diff
@@ -0,0 +1,144 @@
...
+
+// HealthCheckInterval is the time between two health probes of a connected
+// MCPServer. Overridable via MUSTER_ORCHESTRATOR_HEALTH_CHECK_INTERVAL (a Go
+// duration) so the integration test harness can observe the recovery
+// without waiting for production-length ticks.
+var HealthCheckInterval = durationFromEnv("MUSTER_ORCHESTRATOR_HEALTH_CHECK_INTERVAL", 30*time.Second)
+
+// HealthCheckTimeout bounds one probe. A backend that does not answer a ping
+// within it counts as a failed probe.
+const HealthCheckTimeout = 10 * time.Second
```
teemow: Fixed 10 s regardless of the server's `spec.timeout`, which `getRemoteInitContext` honours (default 30 s). A slow backend configured with `timeout: 60` whose ping — or the session re-initialize a ping can trigger — takes 15 s fails three probes, is restarted, reconnects under its 60 s budget and repeats every 90 s. Deriving the probe timeout from `definition.Timeout` (capped) avoids the loop.

internal/services/mcpserver/service.go
```diff
@@ -660,38 +672,66 @@ func (s *Service) GetServiceData() map[string]interface{} {
...
-		s.UpdateHealth(services.HealthUnhealthy)
-		err := fmt.Errorf("MCP client not available")
-		s.emitHealthCheckFailedOnce(err.Error())
-		return services.HealthUnhealthy, err
+		return s.recordHealthCheckFailure(fmt.Errorf("MCP client not available"))
 	}
 
 	// Use MCP ping to check health instead of process checking
 	if pinger, ok := client.(interface{ Ping(context.Context) error }); ok {
 		if err := pinger.Ping(ctx); err != nil {
```
teemow: Any `Ping` error counts as a dead connection. mcp-go turns a JSON-RPC error reply into a Go error, so a backend or proxy that answers ping with `-32601` is restarted every 90 s forever while every `tools/call` succeeds. Also worth a guard: mcp-go v1.0.0 `Client.Ping` is `if c.isModern() { return nil }` (ping was removed in protocol 2026-07-28), so on a modern negotiation every probe passes without any I/O and the feature silently turns off. We pin 2025-11-25 today, but a server echoing 2026-07-28 to the legacy initialize is accepted by `IsValidProtocolVersion`.

teemow (changes requested):
Pushed 3799f348 onto this branch: three behavioral scenarios reproducing the findings below, plus a `pings: true` flag on `test_set_mock_server_outage` so an outage can fail the health probe's pings (default unchanged; the gate now inspects a request body only while it is armed). On the current head they fail at the step that names the defect, so CI is red until the loop is fixed — they are meant as the acceptance criteria for this PR.

| Scenario | Fails today at |
|---|---|
| `mcpserver-health-probe-spares-session-auth-server` | `no-health-check-failure-emitted`: `health check failed: 3 consecutive failures: MCP client not available`, recovery started, recovery failed with 401 |
| `mcpserver-health-probe-spares-per-session-oauth-server` | same events for a per-session OAuth server |
| `mcpserver-health-probe-restart-failure-is-retried` | `reconnected-on-its-own`: stays `failed` after the outage ends; instance log `No retry backoff set for behind-reprogrammed-route, skipping automatic retry` |

`mcpserver-health-probe-restarts-dead-backend`, `mcpserver-remote-backoff-cap` and `mcpserver-restart-request-endpoint-down` still pass with the mock change; `go test ./internal/testing/mock/`, vet and golangci-lint are clean.

The main findings are inline. The one on `health_check.go:43` is the blocker: every forwardToken / tokenExchange / per-session OAuth server with a logged-in session flaps connected → auth_required 90 s after each login in production.

A shape worth considering, because it dissolves several inline points at once: on the third failed probe, close the client and `UpdateState(StateFailed, HealthUnhealthy, err)` with `nextRetryAfter` due now (keep the probe counter; reusing `consecutiveFailures` would inflate the backoff). The existing retry loop, `shouldAttemptRetry`, the aggregator's deregistration on Failed and the reconciler's `applyRetrySchedule`/status then handle the rest: one restart bound instead of two, no in-flight map, no slot wait and no post-slot re-check, and the CR shows `Failed` with `lastError`/`nextRetryAfter` instead of `Connected` with a freshly bumped `lastConnected` for the whole degradation (today `applyStatusFromService` writes exactly that on the health-only reconcile, and `MCPServerStatus` carries no health field). Cost: the restart lands up to `RetryInterval` later, and `core_service_status` loses the connected/unhealthy reading.

Smaller, not inline: `pingOnce` holds the client RLock for the full 10 s against a hung backend, so a `Stop` inside that window blocks `closeClient` — and with it `GetServiceData`/`core_service_status` via `clientInitMutex` — for up to 10 s. The Restart-failure branch, the `ctx.Done` slot wait and the `healthTicker` wiring have no unit coverage (`mockService.restartErr` is there). `MUSTER_ORCHESTRATOR_HEALTH_CHECK_INTERVAL` is missing from docs/how-to/mcp-server-management.md next to its three siblings, and the chart has no way to set any of these env vars, so the 30 s / 3 × 10 s behaviour is fixed on every chart deployment.

internal/orchestrator/health_check.go
fiunchinho: Fixed in c738bbb. CheckHealth returns before probing when the server uses session-level auth (UsesSessionAuth) or holds no shared client, so a server a login synced to Connected is never counted. Both scenarios pass.

internal/services/mcpserver/health_check_test.go
fiunchinho: Rewritten in c738bbb: TestCheckHealth_SkipsServersWithoutASharedClient asserts that a nil client and a forwardToken server are not probed and never counted.

internal/orchestrator/health_check.go
fiunchinho: Fixed in c738bbb. A reconnect that follows failed probes sets reconnectAfterProbe, and while it is set Start schedules a retry for any error, transient or not; the next attempt that reaches the endpoint clears it. TestStart_AfterFailedProbesRetriesAnyError drives a real httptest 404 through it (and pins that a first connect against the same 404 stays unscheduled, as before). Your scenario passes.

internal/orchestrator/health_check.go
fiunchinho: Dissolved by the shape from your summary in c738bbb: the probe path no longer calls Restart. At the threshold the service closes its client and moves to Failed with nextRetryAfter due now, and the retry loop is the only restarter, as before this PR. The pre-existing Start/Stop/Restart overlap between the reconciler and the retry loop is unchanged and listed under 'Not in this PR'.

internal/orchestrator/health_check.go
fiunchinho: Gone with the slot in c738bbb; nothing waits and nothing re-checks.

internal/orchestrator/health_check.go
fiunchinho: Gone in c738bbb: one restart executor, the retry loop with its MaxConcurrentRetries bound.

internal/orchestrator/health_check.go
fiunchinho: Fixed in c738bbb. The orchestrator returns before probing once its context is done, and CheckHealth does not count a probe when the caller's context is done or the error is context.Canceled; the probe's own deadline still counts. The select is gone with the slot.

internal/orchestrator/health_check.go
fiunchinho: Fixed in c738bbb: the probe context comes from getRemoteInitContext, so spec.timeout (default 30s) bounds it; a stdio server gets DefaultStdioInitTimeout. TestCheckHealth_ProbeUsesTheServersTimeout.

internal/services/mcpserver/service.go
fiunchinho: Both fixed in c738bbb. A JSON-RPC error reply (mcp.ErrMethodNotFound, ErrInvalidRequest, ErrInvalidParams, ErrParseError, ErrInternalError) is not counted. A connection whose negotiated version is 2026-07-28 or later is probed with tools/list, since mcp-go's Ping returns nil without I/O there. TestCheckHealth_DoesNotCountAnswersOrCancellations and TestCheckHealth_ProbesModernProtocolWithToolsList.

fiunchinho:
Addressed the review in c738bbb, following the shape from the summary: the orchestrator only probes; at the third failed probe the service closes its client and moves to Failed with nextRetryAfter due now; the existing retry loop reconnects it. Restart slots, in-flight map and post-slot re-check are gone.

Not inline:
- Ticker wiring is covered by TestRetryLoopProbesOnTheHealthTicker; the restart-failure branch no longer exists, and TestFailedProbesFeedTheRetryLoop pins the hand-off.
- MUSTER_ORCHESTRATOR_HEALTH_CHECK_INTERVAL is documented next to its three siblings in docs/how-to/mcp-server-management.md. Chart values for the four variables are listed under 'Not in this PR' since none of them can be set from the chart today.
- pingOnce holding the client RLock for the probe's duration is the pattern every in-flight operation has (a tool call blocks Close the same way); the probe is now bounded by spec.timeout rather than 10s. Left as is.

Your three scenarios pass, as do the four related ones; full suite 234/234 locally; go test -race and golangci-lint clean. PR description updated.

teemow (approved):
Checked c738bbb against the review; this is the shape from the summary and it holds up.

- The orchestrator only probes. `recordHealthCheckFailure` closes the client, sets `nextRetryAfter` due now plus `reconnectAfterProbe`, and moves to Failed; `shouldAttemptRetry` picks it up, `Restart` skips Stop for a Failed service, and `Start` schedules any error while the flag is set. The flag clears on a successful initialize or a 401 (`enterAuthRequired`), and the schedule's `consecutiveFailures` is untouched, so the backoff is not inflated by the probes.
- `UsesSessionAuth()` and a nil client are not probed. The JSON-RPC sentinels survive the real client path: `ping` returns mcp-go's error as is, `listToolsOnce` wraps with `%w`, and `AsError()` returns the sentinels themselves, so `isConnectionFailure` works outside the fake too.
- Ran the seven scenarios from the description on a fresh build of c738bbb (one `--scenario` each, own ports): 7/7 pass. `go test -race` on `internal/orchestrator`, `internal/services/mcpserver`, `internal/aggregator`, `internal/testing/mock`: clean. CI is green on the same head.

Not for this PR: `Stop` leaves `reconnectAfterProbe` set, so a server an operator stops after failed probes keeps the any-error schedule on its next manual start until an attempt reaches the endpoint. Same lifetime `nextRetryAfter` already has.

PR #1196, opened 2026-09-10
fix(metatools): filter_tools echoes a header-declared toolset

### LINKED ISSUE #1194: filter_tools does not echo a header-declared toolset (toolset: null when only X-Muster-Toolset is set)
## Problem

`docs/reference/toolsets.md` and the #1169 contract say the `filter_tools` response "gains `toolset` (echo)". That echo is only produced when the toolset arrives as the `toolset` *argument*. When the toolset is declared by the request header alone — which is how every agent on the platform declares it (`X-Muster-Toolset` rendered by the agent chart) — `filter_tools` answers `"toolset": null`, while `total`/`filtered_count` are scoped and `toolset_unmatched` is filled.

Observed on 5.15.x with `X-Muster-Toolset: preset:read-only` and `filter_tools({"query": "list pods"})`:

```json
{"total": 131, "filtered_count": 5, "truncated": true, "toolset": null, ...}
```

`internal/metatools/handlers.go` (`filterToolsWithOptions`, the `switch` that fills the response): the `argToolset != nil` case sets `resp.Toolset`; the `cat.scoped` case sets only `resp.ToolsetUnmatched`.

Consequence: an agent has no way to learn which toolset bounds it. The only place the toolset is ever visible to the model is the refusal text of `call_tool`. An agent asked "what are you allowed to do?" cannot answer from `filter_tools`, and an operator reading a transcript cannot tell from the discovery calls which toolset was in force.

## Proposed solution

In the `cat.scoped` case, echo the request-declared toolset the same way the argument case does: `resp.Toolset = cat.ts.Raw` (the selectors as declared, e.g. `["preset:read-only"]`). When both header and argument are present the argument case keeps winning (it resolves within the header's toolset), which matches the documented rule. Consider setting `presets` in the header case as well when `include_presets` is not passed, or leave that to `include_presets`; either is fine as long as the docs say which.

## Acceptance criteria

- [x] `filter_tools` with a header-declared toolset and no `toolset` argument returns `toolset` = the header's selectors.
- [x] With both header and argument, `toolset` echoes the argument (unchanged).
- [x] Without any toolset, `toolset` stays absent/null (unchanged).
- [x] A `toolset-*` scenario covers the header-only echo.
- [x] `docs/reference/toolsets.md` states that the echo reflects the header when no argument is given.

Sibling: #1193 (`list_tools` unbounded), found in the same end-to-end check of a `preset:read-only` agent.

## Result

Fixed in #1196 (merged as 54d6d825), released as v5.16.1.

- `filterToolsWithOptions`: the `cat.scoped` case now sets `resp.Toolset = cat.ts.Raw` next to `toolset_unmatched`; the argument case is unchanged and keeps winning when both are present.
- `presets`: stays behind `include_presets` or the `toolset` argument. A header-declared toolset alone does not add them, so header-scoped discovery calls stay as small as unscoped ones. Documented in `docs/reference/toolsets.md` and `docs/reference/mcp-tools.md`; the `filter_tools` tool description mentions the echo.
- `list_core_tools` shares the engine and echoes the header the same way.
- Tests: `TestToolset_FilterToolsEchoesRequestToolset`, scenario `toolset-filter-tools-echo`.

Verified end to end on a local lab with a dev image (`agentlab toolsets-test` and `platform-test` green), and after the Flux roll of v5.16.1 on a production management cluster with a raw MCP request (`initialize` → `tools/call filter_tools {"query":"list pods","limit":3}`, `X-Muster-Toolset: preset:read-only`, no `toolset` argument):

```json
{"total_tools": 450, "total": 131, "filtered_count": 3, "truncated": true, "toolset": ["preset:read-only"], "tools": ["workflow_failing-pods", "workflow_pods-unschedulable", "workflow_pods-pod-pending"]}
```

Same request, same counts as the observation above; `toolset` is now the header instead of `null`.

teemow:
Closes #1194.

## What

`filter_tools` now echoes the toolset the returned tools were resolved within, whichever way it was declared. In `filterToolsWithOptions` the `cat.scoped` case sets `resp.Toolset = cat.ts.Raw` next to `toolset_unmatched`, so a request carrying only `X-Muster-Toolset` gets its selectors back as declared:

```json
{"total": 5, "filtered_count": 5, "toolset": ["preset:read-only"], ...}
```

- header only → `toolset` echoes the header's selectors (new)
- header + argument → `toolset` echoes the argument (unchanged; it resolves within the header's toolset)
- no toolset → `toolset` absent (unchanged)

**`presets` decision:** stays behind `include_presets` or the `toolset` argument. A header-declared toolset alone does not add them, so the discovery calls of an agent whose header is set by its deployment stay as small as unscoped ones. Documented in `docs/reference/toolsets.md`, `docs/reference/mcp-tools.md`, the `filter_tools` tool description and the `FilterToolsResponse` field comments.

`list_core_tools` shares the engine and echoes the header the same way.

## Why

Every agent on the platform declares its toolset by header. Until now `filter_tools` answered with scoped counts and `toolset_unmatched` but no `toolset`, so an agent had no way to learn which toolset bounds it short of a refused `call_tool`, and a transcript did not show which toolset a discovery call ran under.

## Tests

- Unit: `TestToolset_FilterToolsEchoesRequestToolset` (header only incl. unmatched, header + `include_presets`, header + argument, unscoped, `list_core_tools`).
- Scenario: `toolset-filter-tools-echo` (six steps covering the three acceptance criteria plus presets opt-in).
- `make test` (231 scenarios + unit tests with `-race`), `make lint`: green locally.
- End-to-end in the local lab with a dev image: see the follow-up comment.


teemow:
## End-to-end check in the local lab (agentlab)

Dev image built from this branch (`muster:dev-1194-1d1d4330`) swapped into the lab via `platform.devImages.muster`; the running binary reports `muster version dev-1194-1d1d4330`.

- `agentlab platform-test`: all PASS.
- `agentlab toolsets-test`: all PASS (agent-manager toolset requirement, header on the Agent, per-request resolution and refusal, per-server sign-in scoping, portal Tools step and apply path).

Raw MCP request against the lab muster with `X-Muster-Toolset: preset:read-only` and no `toolset` argument (`initialize` → `Mcp-Session-Id` → `tools/call filter_tools {"query":"list pods","limit":3}`), projected:

```json
{
  "total_tools": 59,
  "total": 21,
  "filtered_count": 3,
  "truncated": true,
  "toolset": ["preset:read-only"],
  "tools": ["workflow_lab-cluster-overview", "x_agent-manager_get_agent_status", "core_workflow_list"]
}
```

Before this change the same request answered without `toolset`.

PR #1207, opened 2026-09-11
feat(metatools): describe_tool states that a tool is reached through call_tool

marians:
### What does this PR do?

`describe_tool` now tells the caller that the tool it describes is reachable only through `call_tool`.

The response gained an `invocation` field next to `inputSchema`, naming the call concretely for the tool at hand:

```json
"invocation": "Call it through the call_tool meta-tool: call_tool{\"name\": \"<tool>\", \"arguments\": <object matching inputSchema>}. Tools inside muster are not callable by name directly — only the meta-tools are."
```

That field is the lever that should actually work: it sits in the detail a model reads immediately before calling, rather than in a tool description it skimmed once at session start. The meta-tool's own description — previously "Get detailed information about a specific tool including its input schema" — now says what it returns and points at that field.

The claim is unconditionally true for this handler — `describe_tool` reads `handler.ListTools`, the aggregated catalogue (`core_*`, `x_*`, `workflow_*`), and the aggregator advertises only the meta-tools over MCP, so there is no tool it can describe that a client could call directly.

Agents that look a tool up before calling it are told how to call it. We have had agents issue inner tool names as direct tool calls — which cannot work — and then fail to recover from the resulting error.

The change is additive on the wire: `description`, `inputSchema`, `server`, `kind` and `annotations` are unchanged, and the consumers that parse the response (the `muster` CLI's `get tool`, the REPL `describe`) ignore fields they do not know. Neither prints the new line, which seems right — a human in the REPL *can* call a tool by name there.

### How does it look like?

```json
{
  "name": "x_kubernetes_list_pods",
  "description": "List pods in a namespace.",
  "server": "kubernetes",
  "kind": "tool",
  "annotations": {"readOnlyHint": true},
  "inputSchema": {"type": "object", "properties": {"namespace": {"type": "string"}}},
  "invocation": "Call it through the call_tool meta-tool: call_tool{\"name\": \"x_kubernetes_list_pods\", \"arguments\": <object matching inputSchema>}. Tools inside muster are not callable by name directly — only the meta-tools are."
}
```

No public issue. The trigger is agents calling inner muster tools directly in practice, and not recovering from the error that follows.

Mainly the wording of the `invocation` line — it is prompt surface for every agent on the aggregator, so phrasing is the substance of this PR.

Both review comments are addressed (`da11e459`). One consequence of the shortened tool description is worth a second opinion: the prohibition now reaches a model only through the `invocation` field, i.e. only once it has actually called `describe_tool`. That is the moment that matters for the failure this fixes, but a model that never calls `describe_tool` reads the rule nowhere.

Also still open: whether `filter_tools` with `include_schema=true` should carry the same line, which I deliberately left out of scope.

### Testing

Unit assertions in `internal/metatools` (`FormatToolDetailJSON` and the `describe_tool` handler), plus the `list-tools-paged` scenario's describe step, which now asserts the invocation line verbatim. Ran green: `list-tools-paged`, `toolset-header-filters-catalogue`, `toolset-core-tool-annotations`, `list-core-tools-full-detail`, `filter-tools-discovery-summary`; `go build ./...` and `go vet ./...` clean.

Two things I could not run clean locally, both pre-existing and untouched by this change: `make lint` does not start (golangci-lint built with go1.26 against a 1.27.1 target), and `internal/agent/oauth` fails because those tests bind a fixed port 3000 that another local process held.

Done in this PR: `docs/reference/mcp-tools.md` gained a `describe_tool` section with the example above, and the package doc in `internal/metatools/doc.go` was updated.


docs/reference/mcp-tools.md
```diff
@@ -54,6 +54,26 @@ Use `list_tools` for a bounded look at what is there; use `filter_tools` to *fin
 }
 ```
 
+#### `describe_tool` — the authoritative detail, and how to call it
+
+`describe_tool` returns one tool's full `description` and `inputSchema` together with its
+`server`, `kind` and `annotations`, and an `invocation` line naming the call that runs it.
+Every tool it can describe lives **inside** Muster: an MCP client sees only the meta-tools, so
+issuing `x_kubernetes_list_pods` as a tool call of its own fails. Such a tool is reached
```
marians: The part `so issuing `x_kubernetes_list_pods` as a tool call of its own fails` is too specific. No need to name a tool here that only exists if a certain MCP server is connected with muster.

internal/metatools/provider.go
```diff
@@ -87,7 +87,7 @@ func (p *Provider) GetTools() []api.ToolMetadata {
 		},
 		{
 			Name:        "describe_tool",
-			Description: "Get detailed information about a specific tool including its input schema",
+			Description: "Get the full description, input schema, origin and annotations of one tool in your catalogue. Every tool it describes lives inside muster and is reachable only through call_tool — calling such a name directly, as if it were one of your own tools, always fails. The response's invocation field spells out the call_tool call for the described tool.",
```
marians: Please shorten the description to save tokens:


```suggestion
			Description: "Get the full description, input schema, origin and annotations of one tool. The response's invocation field spells out the call_tool call for the described tool.",
```

docs/reference/mcp-tools.md
marians: Done in da11e459 — the sentence now reads "issuing an aggregated tool's name as a tool call fails", with no tool named.

The example payload below it still uses `x_kubernetes_list_pods`, which I kept deliberately: the `list_tools` example earlier in the same file already uses that name, so the two sections match. Say the word if you want it generic there too.

internal/metatools/provider.go
marians: Applied verbatim in da11e459.

Worth noting what it costs: the description no longer says the tool is *only* reachable through `call_tool`, so that rule now reaches the model solely through the `invocation` field in the response — i.e. only once it has actually called `describe_tool`. That is the moment that matters for the failure we saw, so I think the trade is right, but it does mean a model that never calls `describe_tool` reads the prohibition nowhere.

teemow (approved):
Verified the claim: the catalogue `describe_tool` reads is aggregated server tools plus core tools, meta-tools are never in it, and the aggregator registers only meta-tools on the MCP server — so there is no tool it can describe that a client could call directly. Built the branch fresh, `internal/metatools` tests and the `list-tools-paged` scenario pass against the fresh binary in PATH.

On your open question (the rule now reaches a model only after it calls `describe_tool`): I'd put the one sentence into `call_tool`'s description rather than back into `describe_tool`'s. It currently says only "Execute a tool with the given arguments." — that is the line a model reads when deciding how to execute, once per session, ~15 tokens. Something like: "Execute a catalogue tool by name. Aggregated tools are callable only this way, never by their own name."

Two wording nits, non-blocking:
- `<object matching inputSchema>` sits inside a JSON-shaped snippet; a model may paste the placeholder literally. Prose keeps it safe: `call_tool with {"name": "x", "arguments": {...}}, arguments per inputSchema.`
- Naming "the call_tool meta-tool" as a concept is the right framing, since clients namespace it (`mcp__muster__call_tool` in Claude Code, prefixed on multi-target agentgateway routes). Keep that.

Follow-up worth an issue, out of scope here: the failure is the *error* a direct call gets, and that is mcp-go's own `tool 'x' not found: tool not found` — v1.0.0 has no hook to change it (hooks are observe-only). Answering that call with the invocation note would fix it at the moment it happens, for every model. Agree on leaving `filter_tools include_schema=true` alone.

marians:
Thanks — all three applied in 141def19.

**The rule now lives in `call_tool`'s description**, as you suggested, rather than back in `describe_tool`'s: "Execute a catalogue tool by name. Aggregated tools are callable only this way, never by their own name. A request that declares a toolset (X-Muster-Toolset) can only call tools inside it; anything else is refused naming the toolset." One correction to your reading: it was not only "Execute a tool with the given arguments." — the toolset sentence was already there, so the new sentence is the second of three, before it. That is the better placement anyway, since the toolset rule is a refinement of how a call is bounded, not of what may be called.

**The placeholder is gone.** The line now reads `Call it through the call_tool meta-tool: call_tool with {"name": "x_pager_t01", "arguments": {...}}, arguments per inputSchema.` — your phrasing. Good catch: a JSON-shaped snippet invites pasting whatever sits in the value position.

**Kept "the call_tool meta-tool"** as the framing, for the namespacing reason you give.

Updated with the string: the `list-tools-paged` assertion, the `docs/reference/mcp-tools.md` example, the CHANGELOG entry, and the `call_tool` metadata test, which asserted a substring of the old description.

On the follow-up: I have not opened an issue yet, and have not verified the mcp-go claim myself — I would rather check whether v1.0.0 really offers no way to shape that error before filing it. Say the word and I will.

PR #1289, opened 2026-09-18
fix(filesystem): a status sync never rewrites the definition file; status lives under status/

### LINKED ISSUE #1288: Flaky: oauth-auth-config-change-resets-live-sessions -- a status sync's rewrite of the definition file discards a concurrent spec change
## Problem

Scenario `oauth-auth-config-change-resets-live-sessions` fails in about half of the CI runs of `go-build` since it landed with #1279, on `main` and on tag pipelines alike, and each failure blocks a release until the workflow is rerun from failed. The failing step is always `session-is-back-to-auth-required`: for the whole 30 s wait `auth://status` keeps answering `connected` for `svc`, and `session_auth_reset` is never logged. The step before it, `definition-carries-the-pin`, passes. `oauth-pinned-authorization-server-change-takes-effect` fails the same way now and then, surfacing as the sign-in rate limit.

## Root cause

In filesystem mode the reconciler's status sync is a read-modify-write of the MCPServer's definition file: read the file, apply the status, fsync a temp file, rename it over the original. The connect the scenario causes triggers two status syncs; `test_pin_mcpserver_authorization_server` writes the pin into the same file between their read and their rename, so the rename puts the stale `forwardToken` spec back. `core_mcpserver_get` had read the pin in between (hence `definition-carries-the-pin` passes), but from then on every reconcile pass finds the definition unchanged: no restart, no re-registration, no reset. The fsync on the CI executor's overlayfs holds the window open for about 30 ms; on a host tmpfs it is about 2 ms, which is why the scenario only failed in CI. The `context deadline exceeded` in the failure text is the last `wait_for_state` poll being cut by the wait's own deadline, not a hung read.

## Solution

Shipped in #1289, muster v5.27.6: the filesystem store keeps status under `status/<kind>/<name>.yaml`, at the definition's own relative path; a status sync neither reads nor writes the definition file, a read overlays the recorded status, a `status:` block left in a definition file by an earlier muster is ignored, and status writes no longer fire the change detector watching the definition directories.

## Acceptance criteria

- [x] The mechanism is named in the fix (what blocked the read, what skipped the reset): the status sync's rename over the definition file discarded the pin; nothing blocked the read, it answered `connected` every second.
- [x] `oauth-auth-config-change-resets-live-sessions` passes in a CI-like container over a run count that beats the baseline with margin: baseline on `main` 10 failures in 15 full-suite runs; with #1289 0 failures in 40 runs, two independent sets of 20 ([counts](https://github.com/giantswarm/muster/pull/1289#issuecomment-5735837823), [second set](https://github.com/giantswarm/muster/pull/1289#issuecomment-5735842165)).
- [x] No timing tweak stands in for the fix: no `wait_for_state`, sleep or debounce value changed.

teemow:
Closes #1288

## What

In filesystem mode the reconciler's status sync was a read-modify-write of the MCPServer's (or Workflow's) definition file: read the file, apply the status, fsync a temp file, rename it over the original. A spec written by another process between that read and that rename -- an operator's editor, a GitOps sync, the test harness -- was overwritten with the stale spec, and the reconciler then saw the old definition: the change never happened.

Status now lives under `status/<kind>/<name>.yaml`, at the definition's own relative path, as a document with the status key alone. A status write stats the definition (NotFound when it is gone, never a resurrection) and writes the status file; the definition file is neither read nor written. A read overlays the recorded status on the definition; a `status:` block inside a definition file written by an earlier muster is ignored. Create and update render the definition without a status key; delete removes the status file with the definition. Status writes no longer fire the change detector watching the definition directories, so a status sync no longer schedules the next reconcile pass.

## The flake this fixes

`oauth-auth-config-change-resets-live-sessions`, step `session-is-back-to-auth-required`, about every second `go-build` run since it landed. From the `--debug` log of a failing run in a CI-like container (`--cpus=4 --memory=8g`, overlayfs, `muster test --parallel 20`), times in seconds:

| time | event |
|---|---|
| 34.214 | `svc` connects; two status syncs start and read the definition (still `forwardToken`) |
| 34.215 | `test_pin_mcpserver_authorization_server` writes the pin into the same file |
| 34.216 | `core_mcpserver_get` reads the file: the pin is there (`definition-carries-the-pin` passes) |
| 34.247 | both status syncs rename their copy over the file: the pin is gone, `forwardToken` is back |
| 34.748 .. 63.98 | every reconcile pass: `MCPServer svc is up to date`; no restart, no `session_auth_reset`; `auth://status` keeps answering `connected` |

In the passing run the same two syncs finished at 15.193 and the pin was written at 15.194. The fsync on overlayfs holds the window open for about 30 ms; on a host tmpfs it is about 2 ms, which is why the scenario only failed in CI. The `context deadline exceeded` in the failure text is the last `wait_for_state` poll being cut by the wait's own deadline, not a hung read.

The same loss makes `oauth-pinned-authorization-server-change-takes-effect` error with the sign-in rate limit (its polls of `core_auth_login` against a pin that never took effect); it failed alongside in one baseline run and in CI build 28359.

## Verification

- Unit tests in `internal/client/filesystem`: the definition file stays byte-identical across a status write with an external edit in between; no status key in created or updated definitions; the recorded status is overlaid on get and list; a legacy `status:` block in a definition file is ignored; delete removes the status file; the workflow round trip.
- Baseline on `main` (d7244a78) in the CI-like container, full suite: the scenario failed in 10 of 15 runs.
- This branch, same container, same suite: 0 of 20 runs failed, no scenario at all; a second independent set of 20 runs of the same patch by a peer session is reported in the comments.


teemow:
Verification in the CI-like container (`gsoci.azurecr.io/giantswarm/architect`, `--cpus=4 --memory=8g`, overlayfs `/tmp`, static binary built from the branch, `muster test --parallel 20 --base-port 30000`, the full suite of 263 scenarios per run):

| tree | runs | runs with `oauth-auth-config-change-resets-live-sessions` failed | runs with `oauth-pinned-authorization-server-change-takes-effect` failed | any other failure |
|---|---|---|---|---|
| `main` d7244a78 | 15 | 10 | 1 (together with the first) | 3 readiness timeouts in one run under host load |
| this branch (a7753bec, same patch as the rebased head) | 20 | 0 | 0 | 0 |

A second, independent set of 20 runs of the same patch by a peer session is reported in the next comment. At a 67 % baseline, 20 clean runs happen by luck with probability 0.33^20 ≈ 2e-10.

teemow:
Second, independent verification of the same patch (a7753bec; code-identical to the rebased head f70892b8, patch-id checked), same container recipe, one container per run, 20 sequential full-suite runs: 257 passed / 0 failed in every run (the 6 Kubernetes-mode scenarios skipped without envtest). `oauth-auth-config-change-resets-live-sessions` passed 20 of 20 with `session_auth_reset` logged exactly once per run; `oauth-pinned-authorization-server-change-takes-effect` passed 20 of 20; no other scenario failed in any run. The baseline run of `main` d7244a78 with the same recipe failed the scenario at `session-is-back-to-auth-required` on the first attempt.

Combined: 0 failures in 40 runs of the patch against 10 failures in 15 runs of `main`.

PR #1294, opened 2026-09-19
fix(agent): the bridge and the REPL wait for a slow tool

### LINKED ISSUE #1293: agent --mcp-server: call_tool times out after 30 s regardless of the aggregator's tool timeout
## Problem

`muster agent --mcp-server` (the CLI bridge other clients drive over stdio) forwards `call_tool` through `agent.Client.CallTool`, which uses the client's fixed 30 s timeout (`internal/agent/client.go`, `timeout: 30 * time.Second`). An aggregator tool that legitimately runs longer — the `MCPServer` CRD allows `spec.timeout` up to 300 s, and a registered server may take a minute to answer a read over many upstream calls — fails through the bridge with `transport error: … context deadline exceeded` after exactly 30 s, while the same call succeeds through the aggregator directly. The `--timeout` flag of `muster agent` bounds only the wait for notifications, not a tool call. `CallToolWithTimeout` and `SetTimeout` exist on the client but the bridge never uses them.

## Proposed solution

The bridge's `call_tool` uses a timeout at least as long as the aggregator's largest tool timeout: `--timeout` (or a new `--call-timeout`) applies to tool calls made through the bridge and the REPL, with a default of 300 s; `call_tool` may additionally take an optional `timeout` argument (seconds) that a client such as `platformctl` sets per call. The notification wait keeps its own flag.

## Acceptance criteria

- [ ] A tool that answers after 60 s completes through `muster agent --mcp-server` `call_tool` with default flags.
- [ ] The REPL's `call` uses the same timeout.
- [ ] A unit test covers the bridge's `call_tool` with a slow tool.

teemow:
Fixes #1293.

## What

A tool call made through `muster agent --mcp-server` (`call_tool`), the REPL's `call` or any CLI command that calls a tool ran under the client's 30 s request timeout, while a tool behind the aggregator may legitimately take up to an `MCPServer`'s `spec.timeout` (300 s). `muster agent --timeout` was bound to nothing.

- `agent.Client` gives tool calls a timeout of their own (`DefaultCallTimeout`, 5 min = the aggregator's largest tool timeout; `SetCallTimeout`, `CallTimeout`). The handshake, listings, resources and prompts keep their 30 s (`SetTimeout`). `callToolDirect` is `callToolDirectWithTimeout` with the call timeout; the unused `SetTimeoutForComplexOperations` is gone.
- `muster agent --timeout` (default 5m, same flag, new meaning) sets the client's call timeout for the REPL and the bridge; the REPL runs a command under that timeout instead of a fixed 5 min.
- The bridge's `call_tool` takes an optional `timeout` argument (seconds) that bounds one call in place of `--timeout`; the bridge consumes it (`takeCallTimeout`), it never reaches the aggregator; it is advertised on `call_tool` only, the other meta-tool definitions stay the aggregator's own.
- CLI reference regenerated; `docs/reference/mcp-tools.md` and the CHANGELOG describe the argument.

## How verified

- Unit tests (`internal/agent/server_mcp_timeout_test.go`), mock aggregator whose `call_tool` answers after a delay or when the request's context ends: the default deadline handed to the transport is 5 min (> 30 s); a tool answering after 200 ms completes under a 2 s call timeout; a 5 s tool is cut at a 50 ms call timeout as a tool error, not a hang; `timeout: 1` bounds one call and is stripped from the forwarded arguments; `timeout: 0.05` cuts a slower tool; `"10"`, `0`, `-3`, `true` are refused before anything is forwarded; the `timeout` property is advertised on `call_tool` only; the REPL's command context carries the client's call timeout.
- `go test ./internal/agent/... ./cmd/... ./internal/cli/... ./internal/metatools/...`, `golangci-lint run -E gosec -E goconst` on the changed packages: clean.
- Live, against a local `muster serve` with a mock tool that answers after 60 s: see the comment below.


teemow:
Live proof against a local `muster serve` (filesystem mode) with a streamable-http MCPServer `slowhttp` (`spec.timeout: 120`) whose one tool `wait` answers after 60 s. "before" is a binary built from `main` at 96a11798, "after" this branch.

**Bridge** (`muster agent --mcp-server`, driven over stdio; `+N s` is measured from the moment the request was written):

```
=== after, default flags: call_tool {"name":"x_slowhttp_wait","arguments":{}}
[+  0s] {"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2025-03-26",...,"serverInfo":{"name":"muster-agent","version":"1.0.0"}}}
[+ 62s] {"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"{\"isError\":false,\"content\":[{\"text\":\"{\\\"waited\\\":\\\"60s\\\"}\",\"type\":\"text\"}]}"}]}}

=== after, per-call timeout: call_tool {"name":"x_slowhttp_wait","arguments":{},"timeout":5}
[+  7s] {"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"Meta-tool execution failed: tool call failed: transport error: failed to send request: failed to send request: Post \"http://localhost:38091/mcp\": context deadline exceeded"}],"isError":true}

=== before, default flags
[+ 32s] {"jsonrpc":"2.0","id":2,"result":{"content":[{"type":"text","text":"Meta-tool execution failed: tool call failed: transport error: failed to send request: failed to send request: Post \"http://localhost:38091/mcp\": context deadline exceeded"}],"isError":true}
```

(The request itself was written 2 s after `initialize`, so 62 s is the tool's 60 s, 7 s the 5 s argument, 32 s the old 30 s.)

**REPL** (`muster agent --repl`, `call x_slowhttp_wait`):

```
after:  [08:19:33] Executing tool: x_slowhttp_wait...
        Result:
          "waited": "60s"
before: [08:20:51] Executing tool: x_slowhttp_wait...
        [08:21:21] Tool execution failed: tool call failed: transport error: failed to send request: failed to send request: Post "http://localhost:38091/mcp": context deadline exceeded
```

Two things seen on the way, both outside this change: the bridge against an aggregator that issues no authentication challenge (a plain local `muster serve`) starts in pending-auth mode and serves `authenticate_muster` only, so the proof planted a token for the endpoint in an isolated `HOME`; and a `stdio` MCPServer's tool calls run under the aggregator's `DefaultTimeout` (30 s) regardless of its `spec.timeout` -- the first version of this proof, with the mock tool as a stdio server, got `no answer within the server's timeout of 30s` from the aggregator after the client had stopped cutting the call.

